PR #1, opened 2018-05-17
Added basic functionalities for non-distributed case

shravan97:
### Partially finished functionalities
* Graph Pruning (Still doing some research)
* Mapper function generator (Not a part of this commit)
* Tests (Test environment yet to be setup)


PyTDF/Operation.py
```diff
@@ -0,0 +1,41 @@
...
+        'Count':'a',
+        'Min':'a',
+        'Max':'a',
+        'Mean':'a',
+        'Sum':'a',
+        'Fill':'a',
+        'Take':'a'
+        }
+        
+        try:
```
etejedor: You can replace these 4 lines with:
```python
return operations_dict.get(name)
```

PyTDF/Node.py
```diff
@@ -0,0 +1,74 @@
...
+        if _get_head is None:
+            self._get_head = lambda : self
+        else:
+            self._get_head = _get_head
+        
+        self.operation = operation
+        self.next_nodes = []
+        self.cur_attr = ""
+        self.value = None
+        self.init_refcount = sys.getrefcount(self)
```
etejedor, line 21: So you save the initial count here so that you can use it as the lower value when pruning? Could we first try to understand why the initial value is 4? And why not always?

PyTDF/TDataFrame.py
```diff
@@ -0,0 +1,32 @@
...
+    
+    """
+    The Python equivalent
+    of ROOT C++'s TDataFrame 
+    data structure
+
+    """
+
+    def __init__(self, treename, filelist, npartitions=None):
+        # TODO (shravan97) : Implement different types of constructors
```
etejedor, line 14: We can start with this one, provided that filelist is actually a string (with possibly wildcards), as the real TDataFrame expects

etejedor: Actually this can also be sort of a pythonization of the constructor in C++ that expects a list of files, so we're good

PyTDF/TDataFrame.py
```diff
@@ -0,0 +1,32 @@
...
+
+        super(TDataFrame, self).__init__(None, None)
+        self.filelist = filelist
+        self.treename = treename
+        self.npartitions = npartitions
+        self.chain = ROOT.TChain(treename)
+        self._tdf = ROOT.Experimental.TDataFrame(self.chain)
```
etejedor, line 21: I understand the creation of the real TDataFrame should be moved to the code that calls the mapper? @diocas @bluehood 
In the case of distributed execution, it is DistROOT that constructs the real TDataFrame from a range. If it is a local execution, it will be the code that actually calls the mapper locally.

shravan97: @etejedor Yup certainly. Since I'm yet to commit the mapper, I've added the creation of real TDF object here temporarily. This will be removed when I commit the mapper. And yes, I agree with you on the fact that it'll be created by DistROOT when we go distributed

eguiraud: yes, the pyTDF should not have to build a TDF.

as discussed the mapper function takes a TDF object as argument, replays the booking of transformations and actions, and returns the list of results.

as such the mapper is general enough that it can be passed to a number of different "executors": a distributed executor (`distROOT`) first and foremost, but also a local executor for testing purposes, etc.

PR #5, opened 2018-05-21
Add mapper generator function without DistROOT integration

shravan97:
## Features in this pull request
* A `MapReduce` class which facilitates users to submit or generate mapper and reducer functions
* Mapper function generator, which returns a dictionary of actions and their respective values
* Tests for `MapReduce` class

## Notes
* The commented out code under `def execute_from_graph()` is the way to go about integrating DistROOT (untested though)

## TODO (in the next PR)
* Write a generic reducer function for map reduce
* Integrate DistROOT for distributed execution


eguiraud:
Hi Shravan,
well done, things are really getting into shape.

Let's not worry about the reduce step for now, we need to prepare things on the c++ side to do it efficiently. I think it's perfectly fine, as a first step, to return a list of results and make the whole list accessible from the pyTDF result proxies.

shravan97:
Hi @bluehood ,
The mapper returns a dictionary of results now. I did this so that I can later map the results with the nodes after execution

PyTDF/MapReduce.py
```diff
@@ -0,0 +1,135 @@
...
+
+        if not node.operation: # If root
+            for n in node.next_nodes:
+                self._dfs(n)
+            return
+
+        if node.operation.op_type == "a":
+            self.actions+=1
+            
+            ops_copy.append(node.operation)
```
etejedor: You can move this append outside of the if, since it is done for every case and sub-case

PyTDF/MapReduce.py
```diff
@@ -0,0 +1,135 @@
...
+            
+            new_var_name = "ta"+str(self.actions)
+            self.operations.append((new_var_name, prev, ops_copy))
+            self.action_node_map[new_var_name] = node
+
+        elif node.operation.op_type == "t":
```
etejedor: Why elif? Only actions and transformations right?

PyTDF/MapReduce.py
etejedor: If there are only actions and transformations, you can move out of the if this append, since it is done in all the cases and sub-cases

tests/test_node.py
```diff
@@ -0,0 +1,95 @@
+from PyTDF import *
+import unittest
+
```
etejedor: This file has a different indentation than the previous ones, I would make them the same

PyTDF/MapReduce.py
```diff
@@ -0,0 +1,135 @@
...
+
+        # from DistROOT import DistTree
+        # dTree = DistTree(root_node.file_list, root_node.treename, root_node.npartitions)
+        # reduced_dict = dTree.ProcessAndMerge(self._get_mapper(root_node), self._get_reducer())
+        # for val in reduced_dict:
+        #     self.action_node_map[val].value = reduced_dict[val]
+        
+        pass
+
+    def _get_mapper(self, root_node):
```
etejedor: I would not place this function in a `MapReduce` module. This function should be a utility for you to get a function that, given a TDataFrame, runs the whole graph. I would not even call it `_get_mapper` but rather `get_callable` or something like that.

You will always call this function when triggered by the access to `Proxy` object and, depending on the backend you are using (local, distributed) you will get its result and do whatever you need to do. In the local backend, you will just construct the initial TDataFrame and call the function. In the distributed, you will call ProcessAndMerge. For the moment, I would forget about the latter and about merging and just make it work local. But it is important that we separate things in a way that makes sense conceptually.

shravan97: @etejedor , I thought creating a separate MapReduce class could also facilitate a user to submit their own mapper/reducer functions. But yes, as you said, distributed execution is not yet established. So, I'll change the name of the class and function for this PR

PyTDF/MapReduce.py
shravan97: I thought of eliminating the case in which `node.operation.op_type` is `None`. This would happen only when the operation doesn't match any element in the dictionary of operations

PyTDF/MapReduce.py
shravan97: `ops_copy.append(node.operation)` has to be done inside the if/else statements because, they will get appended to the `operations` array, which additionally depends on `cur` and `prev` variables (they are just the current and previous names of actions/transformations)

PR #7, opened 2018-05-29
Cleaned up code and separated Backend from PyTDF elements

shravan97:
## Coming up in future PRs
- Change callable generator's implementation
- Introduce distributed execution using Spark and DistROOT


PyTDF/Operation.py
```diff
@@ -16,23 +21,28 @@ def __init__(self, name, *args, **kwargs):
         self.op_type = self._classify_operation(name)
 
     def _classify_operation(self, name):
+        if name.lower() in ("range", "take"):
```
eguiraud: why the `lower`? our API is case sensitive

PyTDF/CallableGenerator.py
```diff
@@ -28,7 +27,7 @@ def _dfs(self, node, prev = 't', ops = []):
                 self._dfs(n)
             return
 
-        if node.operation.op_type == OpTypes.ACTION:
+        if node.operation.op_type == node.operation.TYPES.ACTION:
```
eguiraud: conceptually the `CallableGenerator` does depend on the operations, I don't think there is anything wrong in importing and using `OpTypes` directly..?

PyTDF/Operation.py
```diff
@@ -25,20 +22,20 @@ def _classify_operation(self, name):
...
+        'Histo3D':Operation.TYPES.ACTION,
+        'Profile1D':Operation.TYPES.ACTION,
+        'Profile2D':Operation.TYPES.ACTION,
+        'Profile3D':Operation.TYPES.ACTION,
+        'Count':Operation.TYPES.ACTION,
+        'Min':Operation.TYPES.ACTION,
+        'Max':Operation.TYPES.ACTION,
+        'Mean':Operation.TYPES.ACTION,
+        'Sum':Operation.TYPES.ACTION,
+        'Fill':Operation.TYPES.ACTION,
```
eguiraud: why do you prefer this more verbose form?

PyTDF/TDataFrame.py
```diff
@@ -17,15 +17,16 @@ def __init__(self, treename, filelist):
...
+
+        self.Tchain = ROOT.TChain(self.treename)
 
         for f in self.filelist:
-            self.chain.Add(f)
+            self.Tchain.Add(f)
 
-        return ROOT.Experimental.TDataFrame(self.chain)
\ No newline at end of file
+        self._tdf = ROOT.Experimental.TDataFrame(self.Tchain)
```
eguiraud: I think it would be better if you passed treename and filenames to `TDataFrame`'s ctor and let it decide what to do with them under the hood rather than creating the `TChain` by hand.

PyTDF/TDataFrame.py
```diff
@@ -17,15 +17,16 @@ def __init__(self, treename, filelist):
...
         self.filelist = filelist
         self.treename = treename
-        self.chain = ROOT.TChain(treename)
-        self._tdf = self.get_tdf_temp()
 
-    def get_tdf_temp(self):
-        # TODO (shravan97) : shift this to mapper generator
-        # Temporary function to get 
-        # TDF object until mapper is done
+    def create_tdf(self):
```
eguiraud: The creation of the C++ dataframe should not be the job of pyTDF but the backend's.
In distributed executions, for example, the backend will instruct each worker to create its own C++ dataframe.

PyTDF/backend/Local.py
```diff
@@ -0,0 +1,17 @@
...
+    """
+    mapper = generator.get_mapper_callable()
+    generator.root_node.create_tdf()
+
+    output = mapper(generator.root_node._tdf)
+
+    node_map = generator.action_node_map
+
+    for n in node_map:
+        node_map[n].value = output[n]
```
eguiraud: niiiiice :+1:

eguiraud (approved):
Looks good to me, just left some very minor comments.
I'm not commenting the logic of the callable generator and its usage from the executors just yet as we know some changes are expected soon. In any case it looks fine, although probably we will change the local/distributed selection mechanism with something else than an environment variable.

PyTDF/CallableGenerator.py
```diff
@@ -26,7 +27,7 @@ def _dfs(self, node, prev = 't', ops = []):
                 self._dfs(n)
             return
 
-        if node.operation.op_type == "a":
+        if node.operation.op_type == node.operation.TYPES.ACTION:
```
etejedor: Since the enum is defined as a class data member, not an instance data member, it makes more sense to access it through the class (`Operation.TYPES.ACTION`). Also, I would put in lowercase the name of the enum, leaving just the value as uppercase: `Operation.Types.ACTION`.

PyTDF/CallableGenerator.py
```diff
@@ -4,12 +4,13 @@ class CallableGenerator(object):
     mapper abd reducer functions
```
etejedor: Typo: abd --- and

PyTDF/CallableGenerator.py
```diff
@@ -53,21 +54,21 @@ def _dfs(self, node, prev = 't', ops = []):
                 self._dfs(n, *new_tuple)
 
 
-    def get_mapper_callable(self, root_node):
+    def get_mapper_callable(self):
```
etejedor: I suggest we just call this `get_callable(self):`, the concept of mapper is only related to the way some backends will handle this. This is more generic, it is intended to get us a callable to which we can pass an RDF and it will apply all the operations of the graph to it.

PyTDF/CallableGenerator.py
eguiraud: I agree, I liked even more the very first implementation where the enum was part of the `Operation.py` module, so one could do `from Operation import OpTypes` and `OpTypes.ACTION`.

PyTDF/Node.py
```diff
@@ -31,7 +31,7 @@ def _call_handler(self, *args, **kwargs):
         newNode = Node(operation=op, _get_head=self._get_head)
         self.next_nodes.append(newNode)
 
-        if op.op_type == "a":
+        if op.op_type == op.TYPES.ACTION:
```
etejedor: Same comments as above: access with the class and lowercase for the enum type name

PyTDF/Node.py
```diff
@@ -55,20 +54,32 @@ def _dfs(self, node, prev_object= None):
...
         children = []
 
         for n in node.next_nodes:
-            if self._graph_prune_util(n):
+            if self._graph_prune(n):
                 children.append(n)
         node.next_nodes = children
 
-        if len(node.next_nodes) == 0 and (len(gc.get_referrers(node)) <= 3):
+        if not len(node.next_nodes) and (len(gc.get_referrers(node)) <= 3):
```
etejedor: `if not node.next_nodes and` ....

PyTDF/Node.py
```diff
@@ -55,20 +54,32 @@ def _dfs(self, node, prev_object= None):
             self._dfs(n, node.value)
 
     
-    def _graph_prune_util(self, node):
+    def _graph_prune(self, node=None):
```
etejedor: Why do you need the node parameter? Can't you just prune `self`? In the first call, `self` will be the head node, later in the recursion it will be one of the children. This way you do not need the `if not node` code.

PyTDF/Operation.py
```diff
@@ -1,38 +1,45 @@
...
 class Operation(object):
-    
     """
     A Generic representation
     of an operation such as 
     transformations and actions
 
     """
 
+    TYPES = Enum("TYPES", "ACTION TRANSFORMATION")
```
etejedor: `Types` instead of `TYPES`

PyTDF/Operation.py
shravan97: Ok. I'll change this to CamelCase 👍

PyTDF/CallableGenerator.py
shravan97: Yup, makes sense. Initially, I'd moved the CallableGenerator class to inside the backend and then later moved it back outside. I'll change it to a class 👍

PyTDF/Operation.py
shravan97: I'll change this 👍

PyTDF/TDataFrame.py
shravan97: That's a good point. I'll change that 👍

PyTDF/TDataFrame.py
shravan97: This function call is triggered only inside the backend. @bluehood , you'd mentioned earlier that the TDF head node has to be passed around in order for it to work. So, I'm calling this function inside the backend only

PyTDF/CallableGenerator.py
shravan97: Yup. I'll fix this 👍

PyTDF/CallableGenerator.py
shravan97: Thanks. Didn't catch this. I'll fix this 👍

PyTDF/TDataFrame.py
eguiraud: The requirement is that the TDF head node must be in scope when you run the event loop.
You can (should) create the head node in the backend and keep it alive until you ran the event loop.
Note that in distributed executions worker nodes probably will not have a handle to the pyTDF head node so they wouldn't be able to create a C++ TDF head node from it.

shravan97: Also, once the bleeding edge version of ROOT is released, I'll remove this and make it inside backend completely. In that case, I don't have to pass the TDF head node around

PyTDF/Operation.py
etejedor: Yes, you are inside `Operation`, then `Operation` does not need to be there

PyTDF/Operation.py
```diff
@@ -1,38 +1,45 @@
...
+        'Max':Operation.TYPES.ACTION,
+        'Mean':Operation.TYPES.ACTION,
+        'Sum':Operation.TYPES.ACTION,
+        'Fill':Operation.TYPES.ACTION,
         }
         
-        return operations_dict.get(name)
+        op_type = operations_dict.get(name)
+
+        if not op_type:
```
etejedor, line 58: Remember to add here a case for `Take` and `Snapshot`, the message in that case should be different, since they are not invalid, just not supported for now. You can have a separate dictionary for the non-supported operations and throw you error depending on that.

eguiraud: I think it's `Take` and `Foreach`.
`Snapshot` should be available in python afaik.

PyTDF/TDataFrame.py
eguiraud: the requirement that the TDF head node must be in scope when you run the event loop will be there in 6.14 (the upcoming ROOT release) so pyTDF will have to live with it I'm afraid (I didn't manage to get this fixed in time for 6.14).

etejedor: @bluehood point is indeed what we want, the real TDF only lives inside the backend. The local backend will create it locally, distributed backends will create it on the worker nodes for a given range.

PyTDF/Node.py
shravan97: This is a very good point. Thanks ! 👍

PyTDF/backend/Executor.py
```diff
@@ -0,0 +1,33 @@
...
+        """
+        Method to execute the 
+        event-loop from the root
+        node based on the execution
+        environment
+
+        """
+
+        import os
+        env = os.environ.get('PYTDF_ENV')
```
etejedor, line 25: It is fine that we do the choice of the backend based on a variable for now, but at some point we have to make a final decision on what mechanism we use. Other options:
- Different import per backend: `from PyTDF import RDataFrame` for local, `from PyTDF.spark import RDataFrame` for distributed with Spark.
- Additional method: `PyTDF.use('spark')`, default being local.

PyTDF/backend/Local.py
```diff
@@ -0,0 +1,17 @@
+import ROOT
+
+def local_executor(generator):
+    """
+    Execution of the event-loop
+    in local environment
+
+    """
+    mapper = generator.get_mapper_callable()
+    generator.root_node.create_tdf()
```
etejedor: Here you will create the real TDF in place, without going to the generator.

PyTDF/backend/Executor.py
shravan97: Yup, certainly 👍

PyTDF/backend/Local.py
shravan97: This is only until the new version of ROOT is released. In the older version (with TDataFrame instead of RDataFrame), the head node of TDF will have to be kept around for execution of event loop and for `GetValue()` execution with TResultPtr

eguiraud: the requirement is still there in v6.14 (see my comment above).
but the requirement that the head node needs to stay in scope is unrelated to where you create it, right?

shravan97: Yes. That's true. If that's the case, I can manually create the TDF head node here and then assign it to a member in PyTDF object. That'd work right ?

eguiraud: > assign it to a member in PyTDF object

you shouldn't need to: the TDF head node only needs to be alive when running the event loop, which happens inside the backend.
and anyway in the distributed case you could not do it, since the TDF head node would be constructed in the workers, which have no access to the original pyTDF graph.

PyTDF/Operation.py
shravan97: @etejedor , @bluehood , I've created a new list instead 👍

PyTDF/Node.py
```diff
@@ -54,19 +54,16 @@ def _dfs(self, node, prev_object= None):
...
-        for n in node.next_nodes:
-            if self._graph_prune(n):
+        for n in self.next_nodes:
+            if n._graph_prune():
                 children.append(n)
-        node.next_nodes = children
+        self.next_nodes = children
 
-        if not len(node.next_nodes) and (len(gc.get_referrers(node)) <= 3):
+        if (not self.next_nodes) and (len(gc.get_referrers(self)) <= 3):
```
eguiraud: you don't really need these parentheses, but I guess it's a matter of personal preference

PyTDF/TDataFrame.py
```diff
@@ -17,16 +17,5 @@ def __init__(self, treename, filelist):
...
-
-        self.Tchain = ROOT.TChain(self.treename)
-
-        for f in self.filelist:
-            self.Tchain.Add(f)
-
-        self._tdf = ROOT.Experimental.TDataFrame(self.Tchain)
\ No newline at end of file
+        self.TChain = None
+        self.TDF = None
```
eguiraud: do we still need these two variables?

PyTDF/backend/Local.py
```diff
@@ -7,9 +7,17 @@ def local_executor(generator):
...
     """
     mapper = generator.get_mapper_callable()
-    generator.root_node.create_tdf()
+    TChain = ROOT.TChain(generator.root_node.treename)
 
-    output = mapper(generator.root_node._tdf)
+    for f in generator.root_node.filelist:
+        TChain.Add(f)
+
+    TDF = ROOT.Experimental.TDataFrame(TChain)
```
eguiraud: please just pass treename and filenames to TDataFrame

PyTDF/backend/Local.py
```diff
@@ -7,9 +7,17 @@ def local_executor(generator):
...
+
+    generator.root_node.TChain = TChain
+    generator.root_node.TDF = TDF
```
eguiraud: these assignments should not be needed (and would anyway be impossible in distributed executions)

eguiraud (changes requested):

PyTDF/CallableGenerator.py
eguiraud: +1

PyTDF/backend/Local.py
```diff
@@ -0,0 +1,25 @@
...
+
+    output = mapper(generator.root_node.TDF)
+
+    node_map = generator.action_node_map
+
+    for n in node_map:
+        node_map[n].value = output[n]
```
eguiraud: @shravan97 where is the event loop triggered?

PyTDF/Operation.py
eguiraud: @shravan97 I don't see `Snapshot` and `Foreach` anywhere..?
They can both go together with `Range` and `Take` (currently not supported).

shravan97: Yep sure. Will add this :+1:

PyTDF/TDataFrame.py
shravan97: @bluehood , yes. I need to store both those variables. Otherwise, it'd throw the error that we discussed yesterday. Also, storing both of them will keep them in scope

PyTDF/Node.py
shravan97: Yes. It isn't required. I'll remove this in the next commit :+1:

PyTDF/backend/Local.py
shravan97: @bluehood , thanks. I'll do this :+1:

PyTDF/TDataFrame.py
eguiraud: this is my understanding: the c++ TDF needs to be in scope _until the end of the event loop_. since the event loop should run in the backend, the TDF only needs to be in scope in the backend. what am I missing?

PyTDF/backend/Local.py
shravan97: The event loop is actually triggered inside `Proxy.py` , where the user applies `GetValue ()` to an action node

eguiraud: that's too late, it wouldn't work in a distributed execution

PyTDF/CallableGenerator.py
eguiraud: @shravan97 ping

shravan97: Oh yes. I'm sorry. I'll push a commit immediately

PyTDF/backend/Local.py
```diff
@@ -0,0 +1,31 @@
...
+
+    node_map = generator.action_node_map
+
+    event_loop_done = False
+
+    for n in node_map:
+        if not event_loop_done:
+            # Trigger event-loop
+            output[n].GetValue()
+            event_loop_done = True
```
eguiraud: this only does something for the first ouput right? which is correct, but this is a weird way to write it :)

shravan97: So, shall I assign a new variable like `var = output[n]` and then do a `var.GetValue()` then ?

eguiraud: first of all you should be sure that there is always at least one result, but I think that's the case.
given that, you can just do `output[0].GetValue()` to trigger the event loop.

shravan97: But @bluehood , the output of the callable is a dictionary and not a list. If it was a list, that would be good enough. But even in the dictionary, I know that the first key would be `ta1`. So I guess something like `output['ta1'].GetValue()` would be good enough ?

eguiraud: ok ok, `output.values()[0].GetValue()`

eguiraud:
```python
output.values()[0].GetValue() # trigger event loop
```

shravan97: Sounds great. I'll fix this 👍

shravan97: @bluehood , fixed 👍

shravan97:
@bluehood , everything fixed now 👍

eguiraud (approved):
Green light from me if all tests pass :confetti_ball:

PyTDF/CallableGenerator.py
```diff
@@ -26,7 +27,7 @@ def _dfs(self, node, prev = 't', ops = []):
                 self._dfs(n)
             return
 
-        if node.operation.op_type == "a":
+        if node.operation.op_type == node.operation.Types.ACTION:
```
etejedor: As I already pointed out,`Types.ACTION` needs to be accessed as a class data member: `Operation.Types.ACTION`, not as an instance data member as you do here. Please make sure you address our comments, and ask if we are not clear enough!

PyTDF/Node.py
```diff
@@ -31,7 +31,7 @@ def _call_handler(self, *args, **kwargs):
         newNode = Node(operation=op, _get_head=self._get_head)
         self.next_nodes.append(newNode)
 
-        if op.op_type == "a":
+        if op.op_type == op.Types.ACTION:
```
etejedor: Same comment as above for this access.

PyTDF/Node.py
```diff
@@ -55,20 +54,29 @@ def _dfs(self, node, prev_object= None):
...
 
-        for n in node.next_nodes:
-            if self._graph_prune_util(n):
+        for n in self.next_nodes:
+            if n._graph_prune():
                 children.append(n)
-        node.next_nodes = children
+        self.next_nodes = children
+
+        if not self.next_nodes and (len(gc.get_referrers(self)) <= 3):
```
etejedor: You can also remove the parenthesis of the second condition in this `and`: `and len(gc...`

tests/test_operation.py
```diff
@@ -4,16 +4,15 @@
 class ClassifyTest(unittest.TestCase):
     def test_action(self):
         op = Operation("Count")
-        self.assertEqual(op.op_type, "a")
+        self.assertEqual(op.op_type, op.Types.ACTION)
```
etejedor: Same comment for this access

tests/test_operation.py
```diff
@@ -4,16 +4,15 @@
...
 
     def test_transformation(self):
         op = Operation("Define", "c1")
-        self.assertEqual(op.op_type, "t")
+        self.assertEqual(op.op_type, op.Types.TRANSFORMATION)
```
etejedor: And also here

PyTDF/CallableGenerator.py
shravan97: @etejedor , thanks for pointing out. Sorry that I didn't realize this. I'll fix this :+1:

shravan97:
@etejedor , It's fixed now and the tests pass too 👍

PR #14, opened 2018-06-04
Change TDF to RDF and make mapper recursive

### LINKED ISSUE #8: Use recursive callable in CallableGenerator
...and consequently remove the `dfs` function and tests from PyTDF.

### LINKED ISSUE #10: Fix graph pruning to also prune transformations with no children
...and update tests accordingly

### LINKED ISSUE #11: Switch to `PyTDF.use(backend, conf)` as backend selection interface

### LINKED ISSUE #12: Rebase PyTDF on current ROOT master
In particular switch from `TDataFrame` to `RDataFrame`.
@etejedor @diocas @dpiparo what should the rebased version be called? `PyRDF`?

shravan97:
This PR contains the following changes : 
* Implement recursive callable (Fixes #8 )
* Fix graph pruning to recursively remove transformations that are leaf nodes (Fixes #10 )
* Implement `PyRDF.use()` backend selection functionality (Fixes #11 )
* Upgrade supported ROOT version to 6.15 or current master on Github (Fixes #12 )

Coming up in the next PR : 
* Integration tests for PyRDF with reference to https://github.com/root-project/root/blob/master/tree/dataframe/test/*.py


PyRDF/backend/Executor.py
```diff
@@ -12,7 +12,7 @@ def check_env(self):
             return "local"
         return env
 
-    def execute(self, generator):
+    def execute(self, generator, env="local", conf={}):
```
etejedor, line 15: This is more a general comment, but sometimes you include changes in a commit that do not have to do with the commit message. Like this change and the ones below are related to the backend selection and to the synchronization of the final results, that is three different topics in the same commit, which makes it harder for us to review. Please take this into account in future commits!

PyRDF/Proxy.py
```diff
@@ -2,6 +2,9 @@
 from .backend import Executor
 from .CallableGenerator import CallableGenerator
 
+Backend_env = "local"
```
etejedor, line 5: I believe these two variables should not be defined in `Proxy`. `Proxy` just imports them from the entity that manages them. That should be some module called `Backend.py`, which actually could be your current `Executor.py`.
Summary: I would rename `Executor.py` to `Backend.py` and define these two variables there, and import them from `Proxy.py`.

PyRDF/Proxy.py
```diff
@@ -2,6 +2,9 @@
...
+Backend_conf = {}
```
etejedor, line 6: Also I would follow the convention of variables names starting with lowercase.

PyRDF/Proxy.py
etejedor: Just to be fully precise, I would have two variables `name` and `config` as class data members of `Backend`.

PyRDF/Proxy.py
```diff
@@ -25,6 +28,6 @@ def _call_handler(self, *args, **kwargs):
         if not self.action_node.value:
             generator = CallableGenerator(self.action_node._get_head())
             executor = Executor()
-            executor.execute(generator)
+            executor.execute(generator, Backend_env, Backend_conf)
```
etejedor, line 31: Actually, if these variables live in `Backend`, you will not even need to pass them here, you just pass the generator and `Backend` knows what to do with it.

PyRDF/__init__.py
```diff
@@ -1,5 +1,24 @@
...
+    ]
+
+    future_backends = [
+    "dask"
+    ]
+
+    if backend in future_backends:
+        raise NotImplementedError(" This backend environment will be considered in the future !")
+    elif backend in available_backends:
+        Proxy.Backend_env = backend
```
etejedor, line 21: This should be set in `Backend` instead: `Backend.name = backend`

PyRDF/Node.py
```diff
@@ -36,24 +36,6 @@ def _call_handler(self, *args, **kwargs):
...
-        else:
-            ## Execution of the node
-            op = getattr(prev_object, node.operation.name)
-            node.value = op(*node.operation.args, **node.operation.kwargs)
-        
-        for n in node.next_nodes:
-            self._dfs(n, node.value)
-
-    
     def _graph_prune(self):
```
etejedor: `_graph_prune` should not be flagged as internal since it is also called from another module (`CallableGenerator`). It should be then `graph_prune`

PyRDF/Node.py
```diff
@@ -79,4 +61,11 @@ def _graph_prune(self):
             
             return False
 
+        elif self.operation and not self.next_nodes and (self.operation.op_type == Operation.Types.TRANSFORMATION):
```
etejedor: So here we remove a transformation node (to which the user might still have a reference) from the graph. Can this be problematic if the user tries to use that node again from their application?

etejedor: If would add a couple of methods to `Operation.py` so that this `self.operation.op_type == Operation.Types.TRANSFORMATION` becomes this `self.operation.is_transformation()` (same for actions)

PyRDF/CallableGenerator.py
```diff
@@ -66,30 +61,22 @@ def get_callable(self):
 
         self.root_node._graph_prune()
 
-        self.actions = 0
-        self.tfs = 0
-        self.operations = []
-        self.action_node_map = {}
-        self._dfs(self.root_node)
+        def mapper(t, node=None, return_vals=[], return_nodes=[]):
```
etejedor, line 64: Can you elaborate on why this does not work with Spark? Thanks!

PyRDF/CallableGenerator.py
```diff
@@ -66,30 +61,22 @@ def get_callable(self):
...
-
-                ops[tp[0]] = obj
-
-                if tp[0][:2] == "ta":
-                    ret_actions[tp[0]] = obj
+            if not node:
+                node = self.root_node
+            else:
+                t = getattr(t, node.operation.name)(*node.operation.args, **node.operation.kwargs)
+                if node.operation.op_type==Operation.Types.ACTION:
```
etejedor: `if node.operation.is_action()` :smiley:

tests/test_node.py
```diff
@@ -46,50 +43,61 @@ def Count(self):
...
-        self.assertEqual(t.ord_list, [])
+        for n in node.next_nodes:
+            DfsTest.traverse(n, node.value)
 
-    def test_dfs_order_big(self):
-        t = DfsTest.Temp()
+        return prev_object.ord_list
+
+
+    def test_dfs_graph_with_pruning(self):
```
etejedor, line 68: So the question here was: what are `test_dfs_graph_with_pruning` and `test_dfs_graph_without_pruning` testing in `test_node.py` that is not being tested already by `test_mapper_without_pruning` and `test_mapper_with_pruning` in `test_callable_generator.py`?

PyRDF/backend/Executor.py
shravan97: Yes @etejedor . I'd commited files after fixing multiple issues. I shouldve pushed a commit per issue. This is a big mistake. I'll make it a point next time to open a PR per issue :)

PyRDF/Proxy.py
shravan97: @etejedor , it'd work only if I declare those variables as global variables inside `Backend.py`. Class data members will get reset to the default values when you invoke it from a different module. I'd initially kept `Backend_env` and  `Backend_conf` as class data members of Proxy class. If I change the value of these members inside `___init.py__` , it wouldn't work as I'm importing Proxy class separately everywhere else. So, are global variables fine inside `Backend.py` ?

PyRDF/Proxy.py
shravan97: :+1:

PyRDF/__init__.py
shravan97: :+1:

PyRDF/Node.py
shravan97: :+1:

PyRDF/Proxy.py
shravan97: 👍

PyRDF/Node.py
shravan97: This is a great question. @etejedor , It wouldn't work if the user tries to re-use the node which was pruned. It'd throw this error : 
```python
---> 33         return getattr(self.action_node.value, self._cur_attr)(*args, **kwargs)
AttributeError: 'NoneType' object has no attribute 'GetValue'
```
This is also another reason why we should not prune the leaf nodes which are transformations. In the current implementation, the only way for the user to fix this is by redeclaring all the necessary nodes (which were pruned). Shall I change it back to the older implementation then ?
@etejedor , @bluehood ?

shravan97: 👍

PyRDF/CallableGenerator.py
shravan97: @etejedor , it's because, this function (mapper) has a reference to `root`, which is a recursive structure. Since Spark pickles all these details, Pickle keeps complaining this : 
```python
PicklingError: Could not pickle object as excessively deep recursion required.
```
And also, it's return values also contain `Node` object, which is recursive and encounters the same error

PyRDF/CallableGenerator.py
shravan97: 👍

tests/test_node.py
shravan97: @etejedor , this function (`test_dfs_graph_with_pruning`) must be renamed to `test_graph_pruning`. And `test_dfs_graph_without_pruning` will have to be renamed to `test_no_pruning`. My mistake that I forgot to rename these. It's main purpose is to check if the nodes are pruned correctly and this uses a dfs function that I defined in the current test class itself to check the order of nodes. `test_mapper_without_pruning` and `test_mapper_with_pruning` test the mapper function that gets generated with and without graph pruning. This already assumes that graph pruning is correct and doesn't test the pruning at all

PyRDF/Node.py
etejedor: That would be a problem then, I suggest that we do not prune transformations to which the user still has a reference, unless @bluehood has some objection?
It is worth noting that, with this code, transformations with no reference from the user are pruned just like any other node, unless they have children (or grandchildren, etc.) that are still referenced by the user. Is this correct @shravan97 ?

PyRDF/CallableGenerator.py
etejedor: We discussed this with @bluehood yesterday. Since it seems it is the `_get_head` data member function that is causing the circular dependency (and thus the inability of pickle to serialize the graph), the `CallableGenerator` could set to `None` the `_get_head` attribute of the nodes of the graph, so that we prevent the serialization problem afterwards. Note that at this point we do not need `_get_head` anymore, since we already have a reference to the head node.
Also, once we check that this works, we should remove the `_get_head` attribute from transformation nodes, since it is only used from `Proxy`, which gets it from an `Action` node.

tests/test_node.py
etejedor: Isn't this line:
https://github.com/shravan97/PyTDF/blob/6d8ee4cd80a112ba9c7344ff2ebe563148da6108/tests/test_callable_generator.py#L57
precisely testing the pruning in `test_callable_generator.py`?

And this line:
https://github.com/shravan97/PyTDF/blob/6d8ee4cd80a112ba9c7344ff2ebe563148da6108/tests/test_node.py#L83
again testing the pruning in `test_node.py`?

PyRDF/Proxy.py
etejedor: So what about having this:
- A superclass `Backend`, implemented in `Backend.py`, which has a constructor that receives a `config` object and assigns it to an instance data member called `config`. It also has a (virtual) method `execute` with this signature:
```
def execute(callable):
  raise NotImplementedError()
```
- Subclasses (e.g. `Local`, implemented in `Local.py`) which inherit from `Backend` and override the `execute` method with the right behaviour in each case.
- `Proxy` defines a class data member called `backend`, initially `None`
- `PyRDF/__init__.py`, in `use`, constructs a backend object of the right type (e.g. `Local`) with the `config` received from the user (or the default one) and assigns that backend object to `Proxy.backend`.

eguiraud:
Hey guys, there are too many different things happening in this PR, I can't follow :sweat_smile:
I'll just comment on two things:
- regarding the error with the spark deserialization of recursive objects: I think we should, in first approximation, completely ignore it. It requires testing of solutions -- @etejedor and I have one or two to propose -- but this should be done once the local backend is up, running and well tested.
- we should absolutely not prune nodes to which users have references: this is the point of keeping track of the reference counts! Although, if I'm not mistaken, during the last meeting we saw that even transformation nodes which users have no references to are not pruned: those should go away

etejedor:
Agreed with @bluehood that our priority is to have the local backend running and passing all the tests for next week.

PyRDF/Node.py
shravan97: Yes @etejedor . It'd prune the transformations without any references anyway  :+1:

tests/test_node.py
shravan97: @etejedor , the function `test_mapper_with_pruning` is just a test case to make sure that the mapper function works as it is even when some parts of the graph are pruned. It makes sure that the mapper function responds to the changes the graph even after pruning. I'm not very sure if this is compulsory though. I'm also getting a feeling of a slight redundancy here.

Only difference is that, the first test (`test_callable_generator.py`) assumes that the pruning function is correct and checks for generated mapper and the second one (`test_node.py`) assumes that the dfs function (`traverse()`) is correct and only tests for correct pruning.

PyRDF/CallableGenerator.py
shravan97: @etejedor , yes. It sounds like a pretty good explanation to me. I'll reset `_get_head` to `None` 👍

PyRDF/Proxy.py
shravan97: @etejedor , sounds like a good solution to me. One question, where would the `execute()` function of `Local` and `Dist` be called. Inside `execute()` of parent class ?

tests/test_node.py
etejedor: How can `test_callable_generator.py` assume that the pruning function is correct, when there is this check?
https://github.com/shravan97/PyTDF/blob/6d8ee4cd80a112ba9c7344ff2ebe563148da6108/tests/test_callable_generator.py#L59

Don't you end up calling `graph_prune` in the same way you do it here:
https://github.com/shravan97/PyTDF/blob/6d8ee4cd80a112ba9c7344ff2ebe563148da6108/tests/test_node.py#L55

shravan97:
@bluehood , a node which has no user reference is pruned immaterial of whether it an action or a transformation node. If you notice inside `_graph_prune()` function of `Node` class, I'm not checking for action or transformation anywhere (until the one I'd added recently for pruning transformation leaf nodes)

eguiraud:
@shravan97 I might have misunderstood the context, I was replying to @etejedor 's question, "I suggest that we do not prune transformations to which the user still has a reference, unless @bluehood has some objection?"

Anyway, if the logic is the same as before plus the pruning of transformation leaves (with no user references) it should be correct.

eguiraud:
@shravan97 I see what you are saying now. I just re-checked the test that we discussed during last meeting, when we said certain transformation nodes should be pruned but aren't.
They are in fact the ones to which users have references so indeed PyTDF is behaving correctly in the test.

If I interpret the test correctly now, one situation that is not tested (and we wrongly thought it was) is the one of a transformation leaf with no user reference to it.

etejedor:
Yes what @bluehood is suggesting should be tested: not only we need to test that we prune referenceless childless actions, but also referenceless childless transformations.

etejedor:
So we need to test two things:
- A transformation node that is childless and referenceless, and so needs to be pruned.
- An action node whose proxy is referenceless, and so needs to be pruned.
Is this correct @shravan97  ?

PyRDF/Proxy.py
etejedor: No, `execute` would be called from `Proxy` on a `backend` object that has been previously injected in `Proxy` (read above). That `backend` object can be e.g. a `Local`, so the `Local` implementation of `execute` will be invoked. Note that the `execute` of the parent class is simulating a virtual method: it raises a exception when you try to call it.

shravan97:
@bluehood and @etejedor , yes. You're both right. Presently, in the tests,  I'm not checking for transformation nodes which are childless and referenceless getting pruned. I'll add a test case for that :+1:

PyRDF/Proxy.py
shravan97: @etejedor , works :+1:

tests/test_node.py
shravan97: @etejedor , the first check (in `test_callable_generator.py`) is to verify that the mapper traversed the graph in the order that we want despite pruning the graph. In this case `graph_prune` is called here : https://github.com/shravan97/PyTDF/blob/4f93732f059555f123495f3c5ed3eae4d6601f05/PyTDF/CallableGenerator.py#L67

eguiraud:
@shravan97 can we close this PR?

shravan97:
@bluehood , I'm just keeping this PR for reference. I'll myself close this once all individual PRs are merged 👍

etejedor:
@shravan97 notice that even if you close the PR, you don't lose it (you can still see its contents).

shravan97:
@etejedor , yes thanks. Didn't realize this

PR #16, opened 2018-06-06
Add transformation pruning test case and remove dfs, Fixes #10

### LINKED ISSUE #10: Fix graph pruning to also prune transformations with no children

shravan97:
Features in this PR : 
* Rename `_graph_prune` to `graph_prune`
* Add a test case to verify pruning of transformation nodes without user any references
* Remove `_dfs` created earlier as a part of `Node` class

### DIFF
--- a/tests/test_node.py
+++ b/tests/test_node.py
@@ -46,39 +46,32 @@ class DfsTest(unittest.TestCase):
             self.ord_list.append(3)
             return self
 
-    def test_dfs_order_small(self):
-        t = DfsTest.Temp()
+    @classmethod
+    def traverse(cls, node, prev_object= None):
+        """
+        Do a depth-first traversal of the graph from the root
+
+        """
+        if not node.operation:
+            prev_object = DfsTest.Temp()
+            node.value = prev_object
+            node.graph_prune()
+
+        else:
+            ## Execution of the node
+            op = getattr(prev_object, node.operation.name)
+            node.value = op(*node.operation.args, **node.operation.kwargs)
+
+        for n in node.next_nodes:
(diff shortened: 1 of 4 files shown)


tests/test_node.py
```diff
@@ -46,50 +46,79 @@ def Count(self):
...
-        node._dfs(node = node._get_head())
+        n5 = n1.Filter()
+
+        obtained_order = DfsTest.traverse(node = node._get_head())
+
+        reqd_order = [1, 2, 2, 2, 3, 2]
+
+        self.assertEqual(obtained_order, reqd_order)
+
+    def test_dfs_graph_with_pruning_transformations(self):
```
etejedor, line 90: This is great!
I was thinking: can you add [here](https://github.com/shravan97/PyTDF/blob/3235d1e21618df328996dd30de71ea0cc1339ba6/PyRDF/Node.py#L48) a comment explaining how the pruning works?
This is how I see it works:
- For transformations, if they are childless and referenceless from the user side, they are pruned.
- For actions, they are always childless (so that condition check is always true anyway) and they are pruned when their corresponding `Proxy` is referenceless from the user side. This makes the`Proxy` object be garbage collected by Python and, as a consequence, the `Action` node becomes referenceless, and that is what we detect in the pruning algorithm.

This explanation of how it works is not trivial and I think deserves a comment in your code [here](https://github.com/shravan97/PyTDF/blob/3235d1e21618df328996dd30de71ea0cc1339ba6/PyRDF/Node.py#L48).

shravan97: Sure @etejedor

PR #17, opened 2018-06-06
Add backend selection Interface with tests, Fixes #11

### LINKED ISSUE #11: Switch to `PyTDF.use(backend, conf)` as backend selection interface

shravan97:
Features in this PR : 
* Define the `use` function to let the user select appropriate backend with config
* Define base class `Backend` class inside `Backend.py` 
* Implement `Local` child class and `Dist` child class
* Introduce a new class member `backend`, which is set inside the `use` function 
* Call the `execute` function using the `Backend` class instance by passing an instance of `CallableGenerator`
* Write tests for `use` function


etejedor:
As a reference for future PRs, I would place the changes in PyRDF and the test changes in different commits.

PyRDF/backend/__init__.py
```diff
@@ -1 +1,2 @@
-from .Executor import Executor
\ No newline at end of file
+from .Local import Local
```
etejedor, line 1: What's the use of these imports? (and also the ones in `PyRDF/__init__.py`)
Normally this is included to be able to do an import like `from PyRDF.backend import Local` instead of `from PyRDF.backend.Local import Local`, but you do not do this anyway right?

tests/test_backend.py
```diff
@@ -0,0 +1,19 @@
+from PyRDF import *
```
etejedor: Preferably, can you import just what you need?

tests/test_backend.py
```diff
@@ -0,0 +1,19 @@
...
+import unittest
+
+class SelectionTest(unittest.TestCase):
+    def test_local_select(self):
+        use("local")
```
etejedor: I would do `import PyTDF` above and here write `PyTDF.use`, for clarity.

PyRDF/backend/__init__.py
shravan97: @etejedor , I'm using `Local` and `Dist` classes inside `PyRDF/__init__.py`
https://github.com/shravan97/PyTDF/pull/17/files#diff-05dc1f29e7693ee14fa16e4147fb45aaR17

shravan97:
@etejedor , shall I separate the tests from actual code and force push to this branch ?

etejedor:
Yes please @shravan97

tests/test_backend.py
```diff
@@ -1,19 +1,19 @@
-from PyRDF import *
+import PyRDF
 import unittest
 
 class SelectionTest(unittest.TestCase):
     def test_local_select(self):
-        use("local")
-        self.assertIsInstance(Proxy.backend, Local)
+        PyRDF.use("local")
+        self.assertIsInstance(PyRDF.Proxy.backend, PyRDF.Local)
```
etejedor: Well, in these cases you can have `from PyRDF import Proxy, Local, Dist` above and access them as `Proxy.backend`. The case of `use` was different because it was a quite generic name totally unqualified.

shravan97: But @etejedor , it'd be repetitive to do `import PyRDF` and then again `from PyRDF import Proxy, Local, Dist` right ?

shravan97:
@etejedor , I've split the commits now 👍

tests/test_backend.py
etejedor: I do not see why, you would be importing a module and then importing some of its submodules.

shravan97: Ok @etejedor . I'll fix this in a moment 👍

shravan97:
@etejedor , made the change you'd requested about importing from `PyRDF` 👍

PyRDF/Proxy.py
```diff
@@ -25,6 +25,11 @@ def _call_handler(self, *args, **kwargs):
 
         if not self.action_node.value:
             generator = CallableGenerator(self.action_node._get_head())
+
+            if not Proxy.backend:
```
etejedor: To make `Proxy` completely agnostic about the `backend` module, I would move the setting of the default backend to here:
https://github.com/shravan97/PyTDF/blob/186df28db5a2a8bbe20f94c58858c5f8c1af351c/PyRDF/__init__.py#L7

Just by adding this:
```python
# Set default backend
Proxy.backend = Local()
```

etejedor:
Can you please rebase this PR with master?

PyRDF/__init__.py
```diff
@@ -2,4 +2,22 @@
 from .Node import Node
 from .Proxy import Proxy
 from .Operation import Operation
-from .CallableGenerator import CallableGenerator
\ No newline at end of file
+from .CallableGenerator import CallableGenerator
+from .backend import Local, Dist
+
+Proxy.backend = Local({})
```
etejedor: Two comments about this line:
- I would add a default value for `config` in the constructor of `Local` so that we do not need to specify an empty dict here.
- I would perhaps move that constructor (exactly the one in `Local`) to `Backend`, because we will always set the configuration in any backend. Then if we need to do something specific for `Local` or any other backend, we can always write another constructor in those classes, and call the constructor of the superclass from that constructor.

shravan97: @etejedor , sounds good. I'll make these changes 👍

shravan97:
@etejedor , rebase done and this PR is mergeable now 👍

PR #19, opened 2018-06-10
Add multiple constructors

shravan97:
Features in this PR : 
* Add `RDataFrame` constructor support for - 
    - `RDataFrame(int)`
    - `RDataFrame(str, list or str, list or str)`
* Create a new Exception type, i.e., `RDataFrameException`

### DIFF
--- a/PyRDF/RDataFrame.py
+++ b/PyRDF/RDataFrame.py
@@ -1,5 +1,6 @@
 from __future__ import print_function
 from .Node import Node
+import ROOT
 
 class RDataFrame(Node):
     
@@ -8,11 +9,59 @@ class RDataFrame(Node):
     of ROOT C++'s RDataFrame 
     data structure
 
+    Supported constructor arguments :
+    - Number of entries (Int)
+        * For eg., RDataFrame(64)
+    - treename (str), fileslist (str or list of str)
+        * For eg., RDataFrame("myTree", "file.root") or RDataFrame("myTree", ["file1.root", "file2.root"])
+    - treename (str), fileslist (str or list of str), default branches (str or list of str)
+        * For eg., RDataFrame("myTree", ["file1.root", "file2.root"], ["branch1", "branch2"])
+
     """
 
-    def __init__(self, treename, filelist):
-        # TODO (shravan97) : Implement different types of constructors
+    def __init__(self, *args):
 
         super(RDataFrame, self).__init__(None, None)
-        self.filelist = filelist
-        self.treename = treename
+
+        args = list(args) # Make args mutable
+        num_params = len(args)
+
+        for i in range(num_params):
+            if isinstance(args[i], list):
+                args[i] = self._get_vector_from_list(args[i])
+
+        try:
+            ROOT.ROOT.RDataFrame(*args) # Check if the args are correct
+        except TypeError as e:
+            raise RDataFrameException(e, "Error creating the RDataFrame !") from None
+            # 'from None' is to supress the traceback of
+            # the redundant error 'e'
+
+        self.args = args
+
+
+    def _get_vector_from_list(self, arg):
+        """
+        Helper method to convert
+        a python list of strings
+        to a vector
+
+        """
+        reqd_vec = ROOT.std.vector('string')()
+
+        for elem in arg:
+            reqd_vec.push_back(elem)
+
+        return reqd_vec
+
+
+class RDataFrameException(Exception):
+    """
+    A special type of Exception
+    that shows up for incorrect
+    arguments to RDataFrame
+
+    """
+    def __init__(self, exception, msg):
+        super(RDataFrameException, self).__init__(exception)
+        print(msg)
--- a/PyRDF/backend/Local.py
+++ b/PyRDF/backend/Local.py
@@ -10,11 +10,7 @@ class Local(Backend):
         """
         mapper = generator.get_callable()
 
-        filenames_vec = ROOT.std.vector('string')()
-        for f in generator.root_node.filelist:
-            filenames_vec.push_back(f)
-
-        rdf = ROOT.ROOT.RDataFrame(generator.root_node.treename, filenames_vec)
+        rdf = ROOT.ROOT.RDataFrame(*generator.root_node.args)
 
         values, nodes = mapper(rdf)
 
--- a/demos/TDF_demo.ipynb
+++ b/demos/TDF_demo.ipynb
@@ -32,7 +32,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "tdf = RDataFrame(filelist=['CMS_Open_Dataset.root'], treename=\"data\")"
+    "tdf = RDataFrame(\"data\", ['CMS_Open_Dataset.root'])"
    ]
   },
   {


PyRDF/RDataFrame.py
```diff
@@ -8,11 +9,78 @@ class RDataFrame(Node):
     of ROOT C++'s RDataFrame 
     data structure
 
+    Supported constructor arguments :
+    - Integer (Int)
```
diocas: If in the following lines you put the name of the parameter, why is it called "Integer" here, instead of "num_entries"?

PyRDF/RDataFrame.py
```diff
@@ -8,11 +9,78 @@ class RDataFrame(Node):
...
-        self.filelist = filelist
-        self.treename = treename
\ No newline at end of file
+
+        num_params = len(args)
+
+        if num_params == 1:
+            ## If there's only one parameter, check if that's an integer
+            num_entries = args[0]
+            if not isinstance(num_entries, int):
```
diocas: The Python way of doing things ("_easier to ask forgiveness than permission_") says that you should do something like:
```
try:
   num_entries = int(args[0])
except ValueError:
   raise (...)
```

diocas: Should we keep with this mentality, @etejedor ?

PyRDF/RDataFrame.py
```diff
@@ -8,11 +9,78 @@ class RDataFrame(Node):
     of ROOT C++'s RDataFrame 
     data structure
 
+    Supported constructor arguments :
+    - Integer (Int)
+        * For eg., RDataFrame(64)
+    - treename (str), fileslist (str or list of str)
+        * For eg., RDataFrame("myTree", "file.root") or RDataFrame("myTree", ["file1.root", "file2.root"])
+    - treename (str), fileslist (str or list of str), columns (str or list of str)
+        * For eg., RDataFrame("myTree", ["file1.root", "file2.root"], ["column1", "column2"])
```
diocas: (being picky) the RDF constructor calls _filenames_ instead of _filelist_ and  _defaultbranches_ instead of _columns_

tests/test_dataframe_histograms.py
```diff
@@ -0,0 +1,28 @@
+import unittest
+import ROOT, PyRDF
```
diocas: I'm not seeing a reason why this commit is coupled with the new constructors as they are different things. Yes, these tests are possible because you added the new constructors, but I would say that it's still 2 different things (but I will allow it if @etejedor agrees :) )

Also, from what we discussed on last week's meeting, you should provide a test that is an exact replica of what is currently in ROOT. This means, no `tdf = PyRDF.RDataFrame(64)`.
What we discussed was that you should find a way of replacing  ROOT.ROOT.RDataFrame with pyRDF dynamically. From my understanding, this would be to re-use (when this project gets merged into ROOT), the tests from RDataFrame. Some ideas that were discussed (I don't know how/if they work, sorry):
 * create a ROOT module, with the RDF replaced by PyRDF, and prepend it to python path (then when importing ROOT this would be the one being picked, but we were not sure this worked)
 * find a way to replace this globally in the tests configurations.

Can you have a look to see if you find any way of achieving this?
Thanks

diocas (changes requested):

PyRDF/RDataFrame.py
shravan97: Thanks for pointing out @diocas . Will fix this 👍

PyRDF/RDataFrame.py
shravan97: Will fix this 👍

tests/test_dataframe_histograms.py
shravan97: @diocas . Yes. Fair enough that the 2 tests could've been a part of a new PR :) 

And if I remember correctly, we discussed the replacing of `ROOT.ROOT.RDataFrame` for the files under `$ROOTSYS/tutorials/dataframe`. Because, at least one of the tests in this PR require the use of PyROOT's `ROOT.ROOT.RDataFrame`. Like here : https://github.com/shravan97/PyTDF/pull/19/files#diff-7ef345175f2df21f8bef48e12ae5ac46R8

Am I right about this @diocas and other mentors ?

etejedor: I think what was discussed in the end was to just replace the import by something like `from PyRDF import RDataFrame as ROOT.ROOT.RDataFrame`. No big deal though, since in the end PyRDF will become ROOT.ROOT.RDataFrame. What is important is that tests pass.

tests/test_dataframe_histograms.py
```diff
@@ -0,0 +1,28 @@
...
+
+class HistogramsFromRDF(unittest.TestCase):
+    @classmethod
+    def setUp(cls):
+        ROOT.gRandom.SetSeed(1)
```
etejedor: Why is this line repeated in the test?

tests/test_dataframe_misc.py
```diff
@@ -0,0 +1,21 @@
+import unittest
+import ROOT, PyRDF
+
+class MiscTests(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        # load all libs and autoparse
+        df=ROOT.ROOT.RDataFrame(1)
```
etejedor: Do we really need this?

PyRDF/RDataFrame.py
etejedor: Yes @diocas , but what about relying on C++ to tell us about the correctness of the arguments? We could create a real RDF here passing the arguments just for the sake of checking the args:
```python
try:
  ROOT.ROOT.RDataFrame(*args)
except Exception, e:
  raise RDataFrameException("Error creating the RDataFrame. " + str(e))
```

etejedor: Oh ok, I see, you need a treatment of the args before you can pass them to C++...

etejedor: I propose we discuss this tomorrow in the call

shravan97:
@etejedor , there is one more way of doing this as well. Instead of explicitly checking for the 2nd and 3rd args to be a python list, why not I just run a loop over all arguments and convert python lists to vectors immaterial of their position ?

`ROOT.ROOT.RDataFrame` call would anyway take care of improper arguments

etejedor:
Yes @shravan97 , that is what I proposed in our last call. Just go over the args, when you find a list turn it into a vector. The day PyROOT does this for you, we will just remove the conversion.

shravan97:
@etejedor , I've made those changes. Is it okay to keep the `RDataFrameException` class inside `RDataFrame.py` itself though ?

etejedor:
@shravan97 yes it is fine for now, since it is only thrown from inside `RDataFrame.py`

PyRDF/RDataFrame.py
```diff
@@ -8,11 +9,58 @@ class RDataFrame(Node):
...
+
+        for i in range(num_params):
+            if isinstance(args[i], list):
+                args[i] = self._get_vector_from_list(args[i])
+
+        try:
+            ROOT.ROOT.RDataFrame(*args) # Check if the args are correct
+        except TypeError as e:
+            print("Error creating the RDataFrame !")
+            raise RDataFrameException(e) from None
```
etejedor: Can you perhaps construct the `RDataFrameException` with two arguments: the previous exception and a message that you will prepend? This way you can eliminate the `print` from here and do everything inside `RDataFrameException`

shravan97:
@etejedor , done 👍

PR #20, opened 2018-06-10
Test case for recursive pruning of childless and referenceless nodes

shravan97:
Features in this PR : 
* Add a new test case to make sure that childless and reference-less nodes are recursively pruned

### DIFF
--- a/tests/test_node.py
+++ b/tests/test_node.py
@@ -99,10 +99,32 @@ class DfsTest(unittest.TestCase):
         n6 = node.Filter()
 
         n5 = n1.Count()
+        ## This node removes references from the filter
+        ## node defined against n5 above
 
         obtained_order = DfsTest.traverse(node = node._get_head())
 
         reqd_order = [1, 3, 2, 2, 3, 2]
+
+        self.assertEqual(obtained_order, reqd_order)
+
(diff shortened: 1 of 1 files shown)


tests/test_node.py
```diff
@@ -99,13 +99,35 @@ def test_dfs_graph_with_pruning_transformations(self):
...
+        ## This node removes references from the filter
+        ## node defined against n5 above
 
         obtained_order = DfsTest.traverse(node = node._get_head())
 
         reqd_order = [1, 3, 2, 2, 3, 2]
 
         self.assertEqual(obtained_order, reqd_order)
 
+    def test_dfs_graph_with_recursive_pruning(self):
```
diocas, line 111: Hi Shravan, this seems ok. Can you thing of any other scenarios we could test?
Do you think the following scenario is different enough from the ones we already have?
```
(...same code as in this function...)
n7 = n2.Count()
```
This tests that, even though `n2 = None`, it doesn't get cleaned because another node depends on it.
Diogo

shravan97: Yes @diocas . That'd a good test case. Thanks for the suggestion :)

shravan97: @diocas , added a new commit to test that :)

PR #23, opened 2018-06-16
Fix raising RDF error issue for Python 2

shravan97:
Features in this PR : 
* Fix an incompatible `Python2` statement (`raise Exception() from None`) in `RDataFrame`
[This has to be merged **before** running the tests and tutorials using `Python 2`]

### DIFF
--- a/PyRDF/RDataFrame.py
+++ b/PyRDF/RDataFrame.py
@@ -33,9 +33,10 @@ class RDataFrame(Node):
         try:
             ROOT.ROOT.RDataFrame(*args) # Check if the args are correct
         except TypeError as e:
-            raise RDataFrameException(e, "Error creating the RDataFrame !") from None
-            # 'from None' is to supress the traceback of
-            # the redundant error 'e'
+            rdf_exception = RDataFrameException(e, "Error creating the RDataFrame !")
+            rdf_exception.__cause__ = None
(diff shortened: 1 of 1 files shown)


PyRDF/RDataFrame.py
```diff
@@ -33,9 +33,10 @@ def __init__(self, *args):
         try:
             ROOT.ROOT.RDataFrame(*args) # Check if the args are correct
         except TypeError as e:
-            raise RDataFrameException(e, "Error creating the RDataFrame !") from None
-            # 'from None' is to supress the traceback of
-            # the redundant error 'e'
+            rdf_exception = RDataFrameException(e, "Error creating the RDataFrame !")
```
etejedor, line 36: I understand this will still show the error message coming from C++, right? The one saying `None of the overloads succeded`.
Also, could we add a unit test that just checks the construction of the PyRDF? With the right arguments and with wrong arguments (assert exception). With supported and unsupported operations.

JavierCVilla: Indeed, that test could also ensure that no other exception is raised, other than `RDataFrameException`.

shravan97: Sure @etejedor and @JavierCVilla . I'll write a test for this 👍

PR #24, opened 2018-06-18
Add documentation for tests and features

shravan97:
Features in this PR : 
* Added comments for all tests and features in Pydoc format. You can now use `help(...)` to find out stuff about a method or a class

### DIFF
--- a/PyRDF/CallableGenerator.py
+++ b/PyRDF/CallableGenerator.py
@@ -2,18 +2,32 @@ from .Operation import Operation
 
 class CallableGenerator(object):
     """
-    Class that generates
-    mapper and reducer functions
+    Class that generates a callable
+    to parse a PyRDF graph.
+
+    Attributes
+    ----------
+    root_node
+        Head node of a PyRDF graph.
 
     """
     def __init__(self, root_node):
+        """
+        Creates a new `CallableGenerator`.
+
+        Parameters
+        ----------
+        root_node
+            Head node of a PyRDF graph.
+
+        """
         self.root_node = root_node
 
     def _dfs(self, node, prev = 't', ops = []):
         """
         Method that does a depth-first
         traversal and takes note of the 
-        necessary operations to carry out
+        necessary operations to carry out.
 
         """
 
@@ -53,15 +67,41 @@ class CallableGenerator(object):
 
     def get_callable(self):
         """
-        Function that converts a given
-        graph onto a mapper function and 
-        returns the same
+        Converts a given graph into
+        a callable and returns the same.
+
+        Returns
+        -------
+        function
+            The callable that traverses through
+            the given PyRDF graph and executes
+            the operations on all nodes.
 
         """
 
         self.root_node.graph_prune()
 
         def mapper(node_cpp, node_py=None):
+            """
+            The callable that recurses through
+            the PyRDF nodes and executes operations
+            from a starting (PyROOT) RDF node.
+
+            Parameters
+            ----------
+            node_cpp
+                The current state's ROOT CPP node. Initially
+                this should be fed a value.
+
+            Returns
+            -------
+                tuple
+                    This tuple consists of 2 lists, the first one
+                    consisting of output values of action nodes and
+                    the second one of action nodes in the same order
+                    as that of the values list.
+
+            """
             ## TODO : Somehow remove references to any Node object 
             ## for this to work on Spark
 
@@ -69,15 +109,23 @@ class CallableGenerator(object):
             return_nodes = []
 
             if not node_py:
+                # In the first recursive state, just set the
+                # current PyRDF node as the head node
                 node_py = self.root_node
             else:
+                # Execute the current operation using the output of the parent node (node_cpp)
                 node_cpp = getattr(node_cpp, node_py.operation.name)(*node_py.operation.args, **node_py.operation.kwargs)
+
                 if node_py.operation.op_type==Operation.Types.ACTION:
+                    # Collect all action nodes in order to return them
                     return_vals.append(node_cpp)
                     return_nodes.append(node_py)
 
             for n in node_py.next_nodes:
+                # Recurse through children and get their output
                 prev_vals, prev_nodes = mapper(node_cpp, n)
+
+                # Attach the output of the children node
                 return_vals.extend(prev_vals)
                 return_nodes.extend(prev_nodes)
 
@@ -88,7 +136,7 @@ class CallableGenerator(object):
     def get_reducer_callable(self):
         """
         A method that returns a 
-        generalized reducer function 
+        generalized reducer function.
 
         TODO(shravan97)
 
--- a/PyRDF/Node.py
+++ b/PyRDF/Node.py
@@ -5,44 +5,122 @@ import sys, gc
 
 class Node(object):
     """
-    A Class that represents
-    a graph Node
+    A Class that represents a node in RDataFrame
+    operations graph. A Node houses an operation
+    and has references to child nodes. For details
+    on the types of operations supported, try :
+
+    import PyRDF
+    help(PyRDF.Operation)
+
+    Attributes
+    ----------
+    operation
+        The operation that this Node represents. This
+        could be `None`.
+
+    next_nodes
+        A list of `Node` objects which represent the
+        children nodes connected to the current node.
+
+    value
+        The computed value after executing the operation in
+        the current node for a particular PyRDF graph. This
+        is permanently `None` for transformation nodes and
+        the action nodes get a `RResultPtr` after event-loop
+        execution.
 
     """
     def __init__(self, _get_head, operation):
-        
+        """
+        Creates a new `Node` based on the 'operation'.
+
+        Parameters
+        ----------
+        _get_head : function
+            A lambda function that returns the head
+            node of the current graph. This value
+            could be `None`.
+
+        operation : PyRDF.Operation
+            The operation that this Node represents. This
+            could be `None`.
+        """
         if _get_head is None:
+            # Function to get 'head' Node
             self._get_head = lambda : self
         else:
             self._get_head = _get_head
         
         self.operation = operation
         self.next_nodes = []
-        self._cur_attr = ""
+        self._cur_attr = "" # Name of the new incoming operation
         self.value = None
 
     def __getattr__(self, attr):
-        self._cur_attr = attr
+        """
+        Intercepts any call to the current node and dispatches
+        it by means of a call handler.
+
+        Parameters
+        ----------
+        attr : str
+            The name of the operation in the new
+            child node.
+
+        Returns
+        -------
+        function
+            A method to handle an operation call to the
+            current node.
+
+        """
+
+        self._cur_attr = attr # Stores new operation name
         return self._call_handler
 
     def _call_handler(self, *args, **kwargs):
+        # Handles an operation call to the current node and
+        # returns the new node built using the operation call.
 
+        # Create a new `Operation` object for the
+        # incoming operation call
         op = Operation(self._cur_attr, *args, **kwargs)
+
+        # Create a new `Node` object to house the operation
         newNode = Node(operation=op, _get_head=self._get_head)
+
+        # Add the new node as a child of the current node
         self.next_nodes.append(newNode)
 
+        # Return a Proxy object if the new node
+        # is an action node
         if op.op_type == Operation.Types.ACTION:
             return Proxy(newNode)
 
         return newNode
     
     def graph_prune(self):
+        """
+        Prunes nodes from the current PyRDF graph under certain conditions. The current
+        node will be pruned if it has no children and the user application does not hold
+        any reference to it. The children of the current node will get recursively pruned.
+
+        Returns
+        -------
+        True
+            if the current node has to be pruned, False otherwise.
+
+        """
 
         children = []
 
         for n in self.next_nodes:
+            # Select children based on
+            # pruning condition
             if n.graph_prune():
                 children.append(n)
+
         self.next_nodes = children
 
         if not self.next_nodes and len(gc.get_referrers(self)) <= 3:
--- a/PyRDF/Operation.py
+++ b/PyRDF/Operation.py
@@ -3,26 +3,91 @@ from enum import Enum
 
 class Operation(object):
     """
-    A Generic representation
-    of an operation such as 
-    transformations and actions
+    A Generic representation of an operation. The
+    operation could be a transformation or an action.
+
+    Attributes
+    ----------
+    Types
+        A class member that is an Enum of the types
+        of operations supported. This can be ACTION
+        or TRANSFORMATION.
+
+    name
+        Name of the current operation.
+
+    args
+        Variable length argument list for the current
+        operation.
+
+    kwargs
+        Arbitrary keyword arguments for the current
+        operation.
+
+    op_type
+        The type or category of the current
+        operation (ACTION OR TRANSFORMATION).
+
+
+    Supported operations
+    --------------------
+
+    - Define
+    - Filter
+    - Histo1D
+    - Histo2D
+    - Histo3D
+    - Profile1D
+    - Profile2D
+    - Profile3D
+    - Count
+    - Min
+    - Max
+    - Mean
+    - Sum
+    - Fill
+    - Report
+
+
+    Upcoming (future) operations
+    ----------------------------
+
+    - Range
+    - Take
+    - Snapshot
+    - Foreach
 
     """
 
     Types = Enum("Types", "ACTION TRANSFORMATION")
 
     def __init__(self, name, *args, **kwargs):
+        """
+        Creates a new `Operation` for the given name
+        and arguments.
+
+        Parameters
+        ----------
+        name : str
+            Name of the current operation
+
+        *args
+            Variable length argument list for the current
+            operation.
+
+        **kwargs
+            Keyword arguments for the current
+            operation.
+
+        """
         self.name = name
         self.args = args
         self.kwargs = kwargs
         self.op_type = self._classify_operation(name)
 
     def _classify_operation(self, name):
-        """
-        Method to classify the given
-        operation as action or 
-        transformation
-        """
+        # Classifies the given operation as action or
+        # transformation and returns the type.
 
         future_ops = [
         "Range",
--- a/PyRDF/Proxy.py
+++ b/PyRDF/Proxy.py
@@ -3,27 +3,60 @@ from .CallableGenerator import CallableGenerator
 
 class Proxy(object):
     """
-    Proxy provides an interface 
-    between the user and the library
-    to exchange the results of action
-    nodes
+    This class wraps action nodes and instances of Proxy act as
+    futures of the result produced by some action. They implement
+    a lazy synchronization mechanism, i.e., when they are accessed
+    for the first time, they trigger the execution of the whole
+    RDataFrame graph.
+
+    Attributes
+    ----------
+    backend
+        A class member to store a backend object
+        based on the configuration set by the user.
+
+    action_node
+        The action node that the current Proxy
+        instance wraps.
 
     """
-
     backend = None
 
     def __init__(self, action_node):
+        """
+        Creates a new `Proxy` object for a
+        given action node.
 
+        Parameters
+        ----------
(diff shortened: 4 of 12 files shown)


shravan97:
@etejedor and other mentors, do you recommend making the comments under a class or a function definition as Numpy style ?

etejedor:
@shravan97 what do you mean Numpy style?

PyRDF/CallableGenerator.py
```diff
@@ -9,48 +9,6 @@ class CallableGenerator(object):
     def __init__(self, root_node):
         self.root_node = root_node
 
-    def _dfs(self, node, prev = 't', ops = []):
```
etejedor: Why is the removal of this function in this PR?

PyRDF/CallableGenerator.py
```diff
@@ -62,22 +20,39 @@ def get_callable(self):
         self.root_node.graph_prune()
 
         def mapper(node_cpp, node_py=None):
+            """
+            The mapper callable that recurses
```
etejedor: I would not call it mapper in the documentation, actually at some point we should just name this function `callable`, since there will be no concept of mapper and reducer here, and the `get_reducer_callable` will be completely removed.

etejedor: Same comment applies for the docstring of `get_callable` and the one of `CallableGenerator`, please modify those as well.

PyRDF/CallableGenerator.py
```diff
@@ -62,22 +20,39 @@ def get_callable(self):
...
+
+            """
             ## TODO : Somehow remove references to any Node object 
             ## for this to work on Spark
 
             return_vals = []
             return_nodes = []
 
             if not node_py:
+                # In the first recursive state
```
etejedor: Perhaps you can use two lines for this comment instead of 3?

PyRDF/CallableGenerator.py
```diff
@@ -62,22 +20,39 @@ def get_callable(self):
...
+                # just set the current PyRDF
+                # node as the head node
                 node_py = self.root_node
             else:
+                # Execute the current operation using the output of the parent node (node_cpp)
                 node_cpp = getattr(node_cpp, node_py.operation.name)(*node_py.operation.args, **node_py.operation.kwargs)
+
                 if node_py.operation.op_type==Operation.Types.ACTION:
+                    # Collect all action nodes inorder to
```
etejedor: inorder -- in order

etejedor: and the whole comment in one line

PyRDF/CallableGenerator.py
```diff
@@ -62,22 +20,39 @@ def get_callable(self):
...
+                    # return them
                     return_vals.append(node_cpp)
                     return_nodes.append(node_py)
 
             for n in node_py.next_nodes:
+                # Recurse through children and get
```
etejedor: One liner

PyRDF/Node.py
```diff
@@ -5,13 +5,19 @@
...
+    operations graph. A Node houses an operation
+    and has references to child nodes. For details
+    on the types of operations supported, try :
+
+    import PyRDF
+    help(PyRDF.Operation)
 
     """
     def __init__(self, _get_head, operation):
-
```
etejedor: Docstring of the constructor?

PyRDF/Node.py
```diff
@@ -5,13 +5,19 @@
...
+
+    import PyRDF
+    help(PyRDF.Operation)
 
     """
     def __init__(self, _get_head, operation):
-        
+
         if _get_head is None:
+            # Function to get 'head' or 'root' Node
```
etejedor: Let's just call it head node

PyRDF/Node.py
```diff
@@ -22,27 +28,46 @@ def __init__(self, _get_head, operation):
         self.value = None
 
     def __getattr__(self, attr):
-        self._cur_attr = attr
+        # Interception of operation calls
```
etejedor: Make it a proper docstring comment and elaborate a bit more: "Intercepts any call to an operation node and dispatches it by means of a call handler."

PyRDF/Node.py
```diff
@@ -22,27 +28,46 @@ def __init__(self, _get_head, operation):
...
+
+        self._cur_attr = attr # Store new operation name
         return self._call_handler
 
     def _call_handler(self, *args, **kwargs):
```
etejedor: Proper docstring here

PyRDF/Node.py
```diff
@@ -22,27 +28,46 @@ def __init__(self, _get_head, operation):
...
             return Proxy(newNode)
 
         return newNode
     
     def graph_prune(self):
+        """
+        A Function to prune the nodes
+        of a PyRDF graph based on
+        user references
+
```
etejedor: Elaborate a bit more: A node will be pruned if it has no children and the user application does not hold any reference to it.

PyRDF/Operation.py
```diff
@@ -7,8 +7,35 @@ class Operation(object):
...
+    - Take
+    - Snapshot
+    - Foreach
+
     """
 
+    # An Enum Class member to categorize operations
     Types = Enum("Types", "ACTION TRANSFORMATION")
 
     def __init__(self, name, *args, **kwargs):
```
etejedor: docstring for the constructor

PyRDF/Operation.py
```diff
@@ -22,6 +49,7 @@ def _classify_operation(self, name):
         Method to classify the given
         operation as action or 
         transformation
+
```
etejedor: Why this extra line?

PyRDF/Operation.py
```diff
@@ -22,6 +49,7 @@ def _classify_operation(self, name):
...
         """
```
etejedor: It does not apply to this PR but: where are my requested `is_action` and `is_transformation` methods plus their corresponding use throughout the rest of the code to make it simpler? :smile: 
Can you make another PR with those changes?

PyRDF/Proxy.py
```diff
@@ -10,20 +10,25 @@ class Proxy(object):
 
     """
```
etejedor, line 22: I would reword the docstring of this class:
"""
Instances of Proxy act as futures of the result produced by some action.
They implement a lazy synchronization mechanism: when accessed for the first time, they trigger the execution of the whole RDataFrame graph.
"""

PyRDF/Proxy.py
```diff
@@ -10,20 +10,25 @@ class Proxy(object):
...
 
+    # A Class member to store a backend
+    # object based on the configuration
+    # set by the user
     backend = None
 
     def __init__(self, action_node):
```
etejedor: constructor docstring

PyRDF/Proxy.py
```diff
@@ -10,20 +10,25 @@ class Proxy(object):
...
 
         self.action_node = action_node
 
     def __getattr__(self, attr):
+        # Intercept calls on the result
```
etejedor: add a proper docstring

PyRDF/Proxy.py
```diff
@@ -10,20 +10,25 @@ class Proxy(object):
...
+        # of an action node
 
         self._cur_attr = attr
         return self._call_handler
     
     def _call_handler(self, *args, **kwargs):
```
etejedor: add docstring here

PyRDF/backend/Backend.py
```diff
@@ -1,4 +1,10 @@
 class Backend(object):
+    """
+    A Generic Backend class (Interface)
+    that can be extended to define
+    specific 'execute()' functions
+
+    """
 
     def __init__(self, config={}):
```
etejedor: Docstring for both init and execute here

etejedor: Another comment, not for this PR, but for another one: there is the possibilty to specify abstract methods in Python, with the following syntax:
```
@abstractmethod
def execute(self, generator): pass
```
The advantage is that the check is not done at execution time like in the current solution. Can you verify that this works both in Python2 and Python3?

PyRDF/backend/Backend.py
```diff
@@ -1,4 +1,10 @@
 class Backend(object):
+    """
+    A Generic Backend class (Interface)
```
etejedor: """
Base class for RDataFrame backends.
Subclasses need to implement 'execute'.
"""

PyRDF/backend/Local.py
```diff
@@ -2,19 +2,29 @@
...
         """
-        Execution of the event-loop
-        in local environment
+        Overridden abstract method from Backend
+        class to execute operations in local
+        environment
+
         """
-        mapper = generator.get_callable()
+        mapper = generator.get_callable() # Get a callable mapper
```
etejedor: Same comment as before, I would avoid referring to the callable as mapper.

etejedor: Just say callable.

PyRDF/backend/Local.py
```diff
@@ -2,19 +2,29 @@
...
+    that executes the event-loop
+    (operations) locally
+
+    """
 
     def execute(self, generator):
         """
-        Execution of the event-loop
-        in local environment
+        Overridden abstract method from Backend
```
etejedor: """
Execute locally the current graph.
"""

PyRDF/backend/Local.py
```diff
@@ -2,19 +2,29 @@
 from .Backend import Backend
 
 class Local(Backend):
+    """
```
etejedor, line 5: """
Backend that relies on the C++ implementation of RDataFrame to locally execute the current graph.
"""

tests/test_node.py
```diff
@@ -1,36 +1,98 @@
 from PyRDF import *
 import unittest
```
etejedor, line 3: Not all the tests in this file are documented, e.g. `test_dfs_graph_with_pruning_actions`, `test_dfs_graph_with_pruning_transformations`, ... Please make sure all of them are documented.

PyRDF/CallableGenerator.py
shravan97: @etejedor , this function was a part of my previous implementation of Callable Generator. I thought I could utilize this commit for some clean up as well

PyRDF/Operation.py
shravan97: @etejedor , oh yes. I'd created those in #14 , which I'd closed to split the commits. Forgot to make a separate PR during that time. Will one now surely :)

PyRDF/backend/Backend.py
shravan97: @etejedor sure. I'll verify and make a new PR for this 👍

shravan97:
@etejedor , I'll make all the required changes 👍 

Btw, this is what I meant by Numpy style : [https://numpydoc.readthedocs.io/en/latest/format.html#class-docstring](https://numpydoc.readthedocs.io/en/latest/format.html#class-docstring) . Making it Numpy style would make it easier for us to move it to host it on [readthedocs.io](readthedocs.io) as well

PyRDF/CallableGenerator.py
etejedor: Does the commit message say cleanup? :smile: Please separate in into another commit for clarity

shravan97: Yup @etejedor . Fair enough. I'll separate this into another commit 👍

etejedor:
@shravan97 maybe you can use that format for class attributes? The rest of the documentation I would leave it as docstrings of classes and methods for now.
I do not know if we want to have our docs in readthedocs.io, but if we ever want to, is it hard to get it there from docstringed modules?

shravan97:
@etejedor . I just checked out readthedocs.io again and it says that we need to create a separate markdown file for different modules. But then, in case of Numpy style docs, the docstrings would already be in Md format. But yes, I'll do what you suggested. Numpy style for classes and simple docstrings for methods 👍

etejedor:
@shravan97 how are we doing with the comments proposed in this PR?

shravan97:
@etejedor , almost done. I'm just taking some time to make the docs a little more extensive. This should ve done in an hour or two :+1:

shravan97:
@etejedor , this is done now 👍 . I made the docs more consistent for all non test files and wrote them a little extensively

shravan97:
And @etejedor , for remove the `_dfs` function, I'll make a new PR, once this is merged 👍

PyRDF/CallableGenerator.py
```diff
@@ -2,18 +2,32 @@
...
     """
-    Class that generates
-    mapper and reducer functions
+    Class that generates a callable
+    to parse a PyRDF graph.
+
+    Attributes
+    ----------
+    root_node
+        Head node a PyRDF graph.
```
etejedor: of a PyRDF graph?

PyRDF/CallableGenerator.py
```diff
@@ -2,18 +2,32 @@
...
 
     """
     def __init__(self, root_node):
+        """
+        Creates a new `CallableGenerator`.
+
+        Parameters
+        ----------
+        root_node
+            Head node a PyRDF graph.
```
etejedor: of a PyRDF graph?

PyRDF/CallableGenerator.py
```diff
@@ -2,18 +2,32 @@
...
 class CallableGenerator(object):
     """
-    Class that generates
-    mapper and reducer functions
+    Class that generates a callable
+    to parse a PyRDF graph.
+
+    Attributes
+    ----------
+    root_node
```
etejedor, line 10: We might as well call the member `head_node`

PyRDF/CallableGenerator.py
```diff
@@ -2,18 +2,32 @@
...
+        Head node a PyRDF graph.
 
     """
     def __init__(self, root_node):
```
etejedor: and this parameter, also `head_node`, to be consistent with the docs.

PyRDF/CallableGenerator.py
```diff
@@ -53,31 +67,65 @@ def _dfs(self, node, prev = 't', ops = []):
...
-        Function that converts a given
-        graph onto a mapper function and 
-        returns the same
+        Converts a given graph into
+        a callable and returns the same.
+
+        Returns
+        -------
+        function
+            The callable function that traverses
```
etejedor: callable function is perhaps redundant, I would just leave either function or callable.

PyRDF/Node.py
```diff
@@ -5,44 +5,140 @@
...
 
     """
     def __init__(self, _get_head, operation):
-        
+        """
+        Creates a new `Node` based on the 'operation'.
+
+        Parameters
+        ----------
+        _get_head : function
```
etejedor, line 40: Perhaps this parameter should not have the leading `_`, since it has to appear (as parameter) in the documentation.

PyRDF/Node.py
```diff
@@ -5,44 +5,140 @@
...
+            A method to handle an operation call to the
+            current node.
+
+        """
+
+        self._cur_attr = attr # Stores new operation name
         return self._call_handler
 
     def _call_handler(self, *args, **kwargs):
+        """
```
etejedor: This is an internal method, so it should not have docstring attached. The documentation should be in the form of comments.

PyRDF/Node.py
```diff
@@ -5,44 +5,140 @@
...
+        Handles an operation call to the current node.
+
+        Parameters
+        ----------
+        *args
+            Variable length argument list for the incoming
+            operation.
+
+        **kwargs
+            Arbitrary keyword arguments for the incoming
```
etejedor: Why arbitrary?

PyRDF/Operation.py
```diff
@@ -3,25 +3,92 @@
...
+        ----------
+        name : str
+            Name of the current operation
+
+        *args
+            Variable length argument list for the current
+            operation.
+
+        **kwargs
+            Arbitrary keyword arguments for the current
```
etejedor: Arbitrary?

PyRDF/Proxy.py
```diff
@@ -3,27 +3,77 @@
...
+        Handles an operation call to the current node.
+
+        Parameters
+        ----------
+        *args
+            Variable length argument list for the incoming
+            operation.
+
+        **kwargs
+            Arbitrary keyword arguments for the incoming
```
etejedor: Arbitrary

PyRDF/RDataFrame.py
```diff
@@ -5,28 +5,52 @@
...
     
     """
-    The Python equivalent
-    of ROOT C++'s RDataFrame 
-    data structure
-
-    Supported constructor arguments :
-    - Number of entries (Int)
+    The Python equivalent of ROOT C++'s
+    RDataFrame data structure.
```
etejedor: RDataFrame class instead of data structure

PyRDF/RDataFrame.py
```diff
@@ -5,28 +5,52 @@
...
+
+    Attributes
+    ----------
+    args
+        A list of arguments that were provided to
+        construct the RDataFrame object.
+
+    Supported constructor arguments
+    -------------------------------
+    - number of entries (Int)
```
etejedor: lowercase `int`

PyRDF/RDataFrame.py
```diff
@@ -43,9 +67,8 @@ def __init__(self, *args):
 
     def _get_vector_from_list(self, arg):
         """
-        Helper method to convert
-        a python list of strings
-        to a vector
+        converts a python list of strings
```
etejedor: Internal method, docstring should be in the form of comments.
Also uppercase `Converts`.

PyRDF/CallableGenerator.py
shravan97: @etejedor , I meant Python RDataFrame graph. Since this isn't the CPP RDF graph, I made it Python explicitly

PyRDF/CallableGenerator.py
shravan97: @etejedor , can I include this along with the PR to remove `_dfs` ?

That PR will be completely for cleaning up and refactoring variable names. So I guess it'd be better to include this in that PR. Is that okay ?

PyRDF/Node.py
shravan97: And this as well. Can I include this as a part of the clean up PR ?

PyRDF/CallableGenerator.py
shravan97: Shall then change this to " Python RDataFrame graph" then @etejedor ?

PyRDF/Node.py
shravan97: I've written arbitrary because the keyword arguments could be anything. But yes, I think it looks much better without "Arbitrary". I'll just leave it as "Keyword arguments for ..... "

PyRDF/CallableGenerator.py
JavierCVilla: I think he means you missed **of**  right before **a PyRDF**

shravan97: Oh yes. I'm sorry that I didn't notice it. Thank @JavierCVilla 😬

PyRDF/CallableGenerator.py
etejedor: Sure

PyRDF/Node.py
etejedor: Yes sure

PyRDF/Node.py
shravan97: But @etejedor , what if the user tries something like this : 

```python
r = PyRDF.RDataFrame(10)
help(r.Define)
```

shravan97: In that case @etejedor , shall I remove the trailing `_` and change it to `call_handler` ?

etejedor: That is another topic. Ideally `help(r.Define)` should show us the C++ documentation for the real `Define`. That is currently not the case and we should hook in that redirection, but I would leave that aside for now.
I would then just keep `_call_handler` with the underscore, because it is internal. Therefore, its documentation needs to be a comment.

shravan97: Ok sure @etejedor 👍

shravan97:
@etejedor , could you review now ?

PR #26, opened 2018-06-20
Clean up code

shravan97:
Features in this PR : 
* Remove the unwanted `_dfs` function from `CallableGenerator` class
* Variable name changes
    - Change `root_node` -> `head_node` in `CallableGenerator` class and refactor
    - Change `_get_head` -> `get_head` in `Node` class and refactor
    - Change `next_nodes` -> `children` in `Node` class and refactor

### DIFF
--- a/PyRDF/Proxy.py
+++ b/PyRDF/Proxy.py
@@ -57,7 +57,7 @@ class Proxy(object):
         # and returns result of the current action node.
 
         if not self.action_node.value: # If event-loop not triggered
-            generator = CallableGenerator(self.action_node._get_head())
+            generator = CallableGenerator(self.action_node.get_head())
             Proxy.backend.execute(generator)
 
         return getattr(self.action_node.value, self._cur_attr)(*args, **kwargs)
--- a/tests/test_node.py
+++ b/tests/test_node.py
@@ -124,7 +124,7 @@ class DfsTest(unittest.TestCase):
             op = getattr(prev_object, node.operation.name)
             node.value = op(*node.operation.args, **node.operation.kwargs)
 
-        for n in node.next_nodes:
+        for n in node.children:
             DfsTest.traverse(n, node.value)
 
         return prev_object.ord_list
@@ -151,7 +151,7 @@ class DfsTest(unittest.TestCase):
         # Action pruning
         n5 = n1.Filter() # n5 was an action node earlier
 
-        obtained_order = DfsTest.traverse(node = node._get_head())
+        obtained_order = DfsTest.traverse(node = node.get_head())
 
         reqd_order = [1, 2, 2, 2, 3, 2]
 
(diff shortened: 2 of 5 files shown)


shravan97:
@etejedor , I've made one more name change apart from the ones we'd discussed. I changed `next_nodes` to `children` in `Node`

shravan97:
@etejedor , made the `_get_head` change you'd suggested. I've pushed a commit for that. Once, all mentors are okay with this PR, I'll squash all the commits as you'd suggested 👍

JavierCVilla:
Looks good to me

PyRDF/Proxy.py
```diff
@@ -57,7 +57,7 @@ def _call_handler(self, *args, **kwargs):
         # and returns result of the current action node.
 
         if not self.action_node.value: # If event-loop not triggered
-            generator = CallableGenerator(self.action_node.get_head())
+            generator = CallableGenerator(self.action_node._get_head())
```
etejedor: Ok, I just realized two things:
- Why are we calling `_get_head()` here, it is not a function, right? It is a data member.
- Since we need to access the data member `_get_head` from `Proxy`, the name of the data member should be `get_head` (it is not private to `Node`).

So this line should be `generator = CallableGenerator(self.action_node.get_head)`

tests/test_node.py
```diff
@@ -151,7 +151,7 @@ def test_dfs_graph_with_pruning_actions(self):
         # Action pruning
         n5 = n1.Filter() # n5 was an action node earlier
 
-        obtained_order = DfsTest.traverse(node = node.get_head())
+        obtained_order = DfsTest.traverse(node = node._get_head())
```
etejedor: Same here

etejedor: And in the rest of lines in this commit

PyRDF/Proxy.py
shravan97: @etejedor ,  `_get_head` is a function. I remember, we decided to keep it as a lambda function to remove references to root node. And even though `Proxy` accesses it, we aren't exposing it the user right ? It still has to be an internal function right ?

etejedor: Oh yes, that invalidates my comment, it is a data member that is also a function and here we need to call it. But the comment about the privacy still holds: `Proxy` is accessing the data member of `Node ` `_get_head`, which happens to be a function, and then it is calling the data member, so it should be `get_head`.

shravan97: Ok @etejedor . I'm that case, shall I remove the most recent commit and keep only the first commit to avoid redundant commits ?

etejedor: Yes please!

shravan97:
@etejedor , removed the 2nd commit now 👍

PR #28, opened 2018-06-21
Implement `Backend` as an abstract class and write tests

shravan97:
Features in this PR : 
* Implement `Backend` as an abstract class
* Write unit tests for the `Backend` abstract class

### DIFF
--- a/PyRDF/backend/Backend.py
+++ b/PyRDF/backend/Backend.py
@@ -1,4 +1,8 @@
-class Backend(object):
+from abc import ABCMeta, abstractmethod
+
+ABC = ABCMeta('ABC', (object,), {})
+
+class Backend(ABC):
     """
     Base class for RDataFrame backends. Subclasses
     of this class need to implement the 'execute' method.
@@ -26,5 +30,6 @@ class Backend(object):
         """
         self.config = config
 
+    @abstractmethod
     def execute(self, generator):
-        raise NotImplementedError("Incorrect backend environment !")
+        pass
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -38,3 +38,32 @@ class SelectionTest(unittest.TestCase):
 
         with self.assertRaises(NotImplementedError):
             PyRDF.use("dask")
+
+class BackendInitTest(unittest.TestCase):
+    """
+    Tests to ensure that the Backend abstract
+    class cannot be instantiated.
+
+    """
+    def test_backend_init_error(self):
+        """
+        Test case to check that any attempt to
+        instantiate the `Backend` abstract class
+        results in a `TypeError`.
+
+        """
+        with self.assertRaises(TypeError):
+            PyRDF.backend.Backend.Backend()
+
+    def test_subclass_without_method_error(self):
+        """
+        Test case to check that creation of a
+        subclass without implementing `execute`
+        method throws a `TypeError`.
+
+        """
+        class TestBackend(PyRDF.backend.Backend.Backend):
+            pass
+
+        with self.assertRaises(TypeError):
+            TestBackend()


shravan97:
@etejedor  and mentors. Any more tests to be added ?

tests/test_backend.py
```diff
@@ -38,3 +38,19 @@ def test_future_env_select(self):
...
+    """
+    def test_backend_init_error(self):
+        """
+        Test case to check that any attempt to
+        instantiate the `Backend` abstract class
+        results in a `TypeError`.
+
+        """
+        with self.assertRaises(TypeError):
+            backend = PyRDF.backend.Backend.Backend()
```
etejedor: Small thing: you do not need to even create a variable here, this line can just be `PyRDF.backend.Backend.Backend()`.

I do not think you need any other test. In the call this morning we mentioned a test that creates a subclass of `Backend` without the `execute` method and is expected to fail, but that would be testing Python, not really our code.

tests/test_backend.py
```diff
@@ -38,3 +38,19 @@ def test_future_env_select(self):
...
         with self.assertRaises(NotImplementedError):
             PyRDF.use("dask")
+
+class BackendInitTest(unittest.TestCase):
+    """
+    Tests to ensure that the Backend abstract
+    class cannot be instantiated.
+
+    """
+    def test_backend_init_error(self):
```
JavierCVilla: @shravan97 We should also test a subclass of `Backend` which does not implement `execute`, since this is the real case we want to avoid. For that you can just create a dummy class on the test itself before the `assert`.

tests/test_backend.py
etejedor: Or after thinking about @JavierCVilla 's comment, perhaps we can just add that last test to double check the `@abstractmethod` decorator is not removed from the base class.

JavierCVilla: I had not seen @etejedor 's review, so I removed mine. Basically, I was just proposing to test not only `Backend` but a subclass of `Backend`, which might be just a dummy class created in the same test, without defining its `execute` method.

shravan97: @JavierCVilla and @etejedor sounds good. I'll that test too. But there's a small problem before that. I just found that python 2 doesn't have `ABC` class. So, I'm finding solutions common to both

JavierCVilla: @shravan97 I think Python 2 also has an `abc` builtin module, have a look at this [link](https://docs.python.org/2/library/abc.html). Could you try using `__metaclass__` and `ABCMeta` instead?

shravan97: @JavierCVilla , yes. I tried this. In this case only Python2 threw me an error on instantiation and Python3 didn't. But I found another fix for that here : [https://stackoverflow.com/questions/35673474/using-abc-abcmeta-in-a-way-it-is-compatible-both-with-python-2-7-and-python-3-5](https://stackoverflow.com/questions/35673474/using-abc-abcmeta-in-a-way-it-is-compatible-both-with-python-2-7-and-python-3-5)

shravan97:
@etejedor and @JavierCVilla , added a couple of commits now. It works with both Python 2 and 3 now 👍

PyRDF/backend/Backend.py
```diff
@@ -1,6 +1,7 @@
-from abc import ABC, abstractmethod
+import abc, six
```
etejedor: This is great, I have a couple of comments:
- We can trust the `six` module to be always there, right? If I install Python, it comes with it out of the box?
- Can we do `import six` and `from abc import ABCMeta, abstractmethod`, so that we do not need to prepend `abc` below?

etejedor:
@shravan97 after addressing my last comment, please reorganize the commits of this PR in just two: one for the changes in the code and one for the tests.

PyRDF/backend/Backend.py
shravan97: @etejedor , I have the same doubt about `six` module. It's available on my local and SWAN (without explicitly installing). But for safer side, I think I'll switch to something else. Here's another workaround for this : [https://gist.github.com/alanjcastonguay/25e4db0edd3534ab732d6ff615ca9fc1](https://gist.github.com/alanjcastonguay/25e4db0edd3534ab732d6ff615ca9fc1)

shravan97: And yes. I'll fix that import 👍

shravan97:
Sure @etejedor . I'll re-organize the commits 👍

PyRDF/backend/Backend.py
JavierCVilla: @shravan97 @etejedor I think `ABC = abc.ABCMeta('ABC', (object,), {})` is the way to go. `six` is an external python module, so we would need to deal with an external dependency. It is present in SWAN because we do provide it on purpose.

shravan97: Yup. I too feel so @JavierCVilla . Shall I finalize and push the commits @etejedor ?

shravan97: @etejedor and @JavierCVilla , I've pushed the 2 commits now 👍

PR #30, opened 2018-06-23
Change Backend implementation to check supported operations

shravan97:
Features in this PR : 
* Add `is_supported` function in `Backend` class to check if an operation is supported
* Add a `supported_operations` and `future_operations` dictionary to `Backend` class
* Remove `Proxy.backend` and create a global named `current_backend` inside `__init__.py` so that all modules inside `PyRDF` can access them (in this case `Proxy` and `Operation`, but may be more in the future)
* Add/Modify tests for the above features

### DIFF
--- a/PyRDF/Node.py
+++ b/PyRDF/Node.py
@@ -87,6 +87,11 @@ class Node(object):
         # Handles an operation call to the current node and
         # returns the new node built using the operation call.
 
+        from . import current_backend
+        # Check if the current operation is supported by
+        # the backend
+        current_backend.check_supported(self._cur_attr)
+
         # Create a new `Operation` object for the
         # incoming operation call
         op = Operation(self._cur_attr, *args, **kwargs)
--- a/PyRDF/Operation.py
+++ b/PyRDF/Operation.py
@@ -89,16 +89,6 @@ class Operation(object):
         # Classifies the given operation as action or
         # transformation and returns the type.
 
-        future_ops = [
-        "Range",
-        "Take",
-        "Snapshot",
-        "Foreach"
-        ]
-
-        if name in future_ops:
-            raise NotImplementedError("This operation ({}) isn't supported yet ! ".format(name))
-
         ops = Operation.Types
 
         operations_dict = {
--- a/PyRDF/__init__.py
+++ b/PyRDF/__init__.py
@@ -5,19 +5,21 @@ from .Operation import Operation
 from .CallableGenerator import CallableGenerator
 from .backend import Local, Dist
 
-Proxy.backend = Local()
+current_backend = Local()
 
-def use(backend, conf = {}):
+def use(backend_name, conf = {}):
 
     future_backends = [
     "dask"
     ]
 
-    if backend in future_backends:
+    global current_backend
+
+    if backend_name in future_backends:
         raise NotImplementedError(" This backend environment will be considered in the future !")
-    elif backend == "local":
-        Proxy.backend = Local(conf)
-    elif backend == "spark":
-        Proxy.backend = Dist(conf)
+    elif backend_name == "local":
+        current_backend = Local(conf)
+    elif backend_name == "spark":
+        current_backend = Dist(conf)
     else:
         raise Exception(" Incorrect backend environment \"{}\"".format(backend))
--- a/PyRDF/backend/Backend.py
+++ b/PyRDF/backend/Backend.py
@@ -14,6 +14,29 @@ class Backend(ABC):
         backend.
 
     """
+    supported_operations = [
+        'Define',
+        'Filter',
+        'Histo1D',
+        'Histo2D',
+        'Histo3D',
+        'Profile1D',
+        'Profile2D',
+        'Profile3D',
+        'Count',
+        'Min',
+        'Max',
+        'Mean',
+        'Sum',
+        'Fill',
+        'Report',
+        'Range',
+        'Take',
+        'Snapshot',
+        'Foreach',
+        'Reduce',
+        'Aggregate'
+        ]
 
     def __init__(self, config={}):
         """
@@ -29,6 +52,25 @@ class Backend(ABC):
 
         """
         self.config = config
+
+    def check_supported(self, operation_name):
+        """
+        Checks if a given operation is supported
+        in the given backend.
+
+        Parameters
+        ----------
+        operation_name
+            Name of the operation to be checked.
+
+        Returns
+        -------
+        Operation.Types
+            The type of the current operation.
+
+        """
+        if operation_name not in self.supported_operations:
+            raise Exception("Invalid operation \"{}\"".format(operation_name))
 
     @abstractmethod
     def execute(self, generator):
--- a/PyRDF/backend/Local.py
+++ b/PyRDF/backend/Local.py
@@ -13,6 +13,33 @@ class Local(Backend):
         backend.
 
     """
+    def __init__(self, config={}):
+        """
+        Creates a new instance of the
+        Local implementation of `Backend`.
+
+        Parameters
+        ----------
+        config (optional)
+            The config object for the required
+            backend. The default value is an
+            empty Python dictionary `{}`.
+
+        """
+        super().__init__(config)
+        operations_not_supported = [
+        'Take',
+        'Snapshot',
+        'Foreach',
+        'Reduce',
+        'Aggregate'
+        ]
+        if ROOT.ROOT.IsImplicitMTEnabled():
+            # Add `Range` operation only if multi-threading
+            # is disabled.
+            operations_not_supported.append('Range')
+
+        self.supported_operations = [op for op in self.supported_operations if op not in operations_not_supported]
 
     def execute(self, generator):
         """
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -1,6 +1,5 @@
-import PyRDF
+import PyRDF, unittest, ROOT
 from PyRDF import Proxy, Local, Dist
-import unittest
 
 class SelectionTest(unittest.TestCase):
     """
@@ -17,7 +16,7 @@ class SelectionTest(unittest.TestCase):
         """
 
         PyRDF.use("local")
-        self.assertIsInstance(Proxy.backend, Local)
+        self.assertIsInstance(PyRDF.current_backend, Local)
 
     def test_dist_select(self):
         """
@@ -27,7 +26,7 @@ class SelectionTest(unittest.TestCase):
         """
 
         PyRDF.use("spark")
-        self.assertIsInstance(Proxy.backend, Dist)
+        self.assertIsInstance(PyRDF.current_backend, Dist)
 
     def test_future_env_select(self):
         """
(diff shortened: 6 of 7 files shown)


shravan97:
@etejedor @diocas @bluehood @JavierCVilla @dpiparo , the implementation in this PR is still up for debate. I guess it's better to continue work on this PR after we discuss the best implementation during our meeting this week

etejedor:
@shravan97 thanks for the PR, I believe it is worth to trigger the discussion already before the meeting. I'm commenting on your changes.

PyRDF/Operation.py
```diff
@@ -88,42 +89,8 @@ def __init__(self, name, *args, **kwargs):
...
-        "Range",
-        "Take",
-        "Snapshot",
-        "Foreach"
-        ]
-
-        if name in future_ops:
-            raise NotImplementedError("This operation ({}) isn't supported yet ! ".format(name))
-
-        ops = Operation.Types
```
etejedor: `Operation.Types`, `ops` and `operations_dict` should stay here, in `Operation`.

PyRDF/Operation.py
```diff
@@ -88,42 +89,8 @@ def __init__(self, name, *args, **kwargs):
     def _classify_operation(self, name):
         # Classifies the given operation as action or
         # transformation and returns the type.
-
-        future_ops = [
```
etejedor: `future_ops` can perhaps go away from `Operation`, and be part of the check for the support in the corresponding `Backend`. Or even disappear completely (we just make a distinction between supported and not supported, we do not talk about future plans).

PyRDF/Operation.py
```diff
@@ -88,42 +89,8 @@ def __init__(self, name, *args, **kwargs):
...
-        'Fill':ops.ACTION,
-        'Report':ops.ACTION
-        }
-        
-        op_type = operations_dict.get(name)
-
-        if not op_type:
-            raise Exception("Invalid operation \"{}\"".format(name))
-        return op_type
+        from . import current_backend
```
etejedor: I think these two lines would not even be necessary. `Operation` should stay as a class that contains **all**  the operations and their types, it does not need to care whether the operation is supported in a backend or not, it should not contact the backend. That check should be done outside of `Operation`: first you ask the backend if it supports that operation, and then you ask `Operation` to classify that operation for you, in two steps.

PyRDF/__init__.py
```diff
@@ -5,19 +5,21 @@
 from .CallableGenerator import CallableGenerator
 from .backend import Local, Dist
 
-Proxy.backend = Local()
+current_backend = Local()
```
etejedor, line 8: Good, but I would change the name of this variable to just `backend`

PyRDF/__init__.py
```diff
@@ -5,19 +5,21 @@
...
 
     future_backends = [
     "dask"
     ]
 
+    global current_backend
+
     if backend in future_backends:
         raise NotImplementedError(" This backend environment will be considered in the future !")
     elif backend == "local":
```
etejedor: and I would change the name of this variable to `backend_name`

PyRDF/backend/Backend.py
```diff
@@ -14,6 +15,30 @@ class Backend(ABC):
         backend.
 
     """
+    supported_operations = {
```
etejedor: You do not actually need the classification of the operations here, it is `Operation` the module that has that information. You just need to state what you support.

PyRDF/backend/Backend.py
```diff
@@ -14,6 +15,30 @@ class Backend(ABC):
...
+        'Count':Operation.Types.ACTION,
+        'Min':Operation.Types.ACTION,
+        'Max':Operation.Types.ACTION,
+        'Mean':Operation.Types.ACTION,
+        'Sum':Operation.Types.ACTION,
+        'Fill':Operation.Types.ACTION,
+        'Report':Operation.Types.ACTION
+        }
+
+    future_operations = [
```
etejedor: Like I commented before, this could even go away.

shravan97: @etejedor , can we keep the `future_operations` variable to show the operations that ROOT C++ RDF supports, but not `PyRDF` ?

PyRDF/backend/Backend.py
```diff
@@ -30,6 +55,32 @@ def __init__(self, config={}):
         """
         self.config = config
 
+    def is_supported(self, operation_name):
```
etejedor: You need to call this from `Node`, when the user just called something on one of your nodes, [here](https://github.com/shravan97/PyRDF/blob/master/PyRDF/Node.py#L92). You do the check in two steps, first you check if the operation name you received `is_supported` by your current backend, and only if that check passes you create the `Operation` object, which will contain the classification.

PyRDF/__init__.py
shravan97: @etejedor , `current_backend` is the same global variable that I'd declared outside. I have to use the `global` keyword to access it. Otherwise, the use function consider it as a variable in its scope. So, shall I rename this to `backend` itself as you'd suggested then ?

shravan97: And also @etejedor , there would be an issue in importing `backend` if I change the name of the global variable to backend. There would be a conflict between this global variable and the sub-package named backend

PyRDF/backend/Backend.py
shravan97: Yes @etejedor . This is a good idea

shravan97:
@etejedor , changed as per your suggestions. `Backend` is now agnostic of `Operation` and `Node` calls the `is_supported` method. Once you're happy with the implementation, let me know and I'll rebase commits in this PR to 2 commits 👍

PyRDF/__init__.py
etejedor: Then I would change `backend` to `backend_name` and keep `current_backend` as it is.

PyRDF/backend/Backend.py
etejedor: I think it is not even worth it, but I would let the others comment

PyRDF/CallableGenerator.py
```diff
@@ -1,5 +1,3 @@
-from .Operation import Operation
```
etejedor: Watch out, this is already in, please rebase with current master

shravan97: Yup @etejedor . Once you're happy with the other changes, I'll rebase surely 👍

PyRDF/__init__.py
shravan97: 👍

PyRDF/backend/Backend.py
shravan97: Ok @etejedor . I'll remove it for now then 👍

PyRDF/Node.py
```diff
@@ -87,6 +87,11 @@ def _call_handler(self, *args, **kwargs):
         # Handles an operation call to the current node and
         # returns the new node built using the operation call.
 
+        from . import current_backend
+        # Check if the current operation is supported by
+        # the backend
+        current_backend.is_supported(self._cur_attr)
```
etejedor: Can we call it `check_supported`? `is_supported` looks like that it should return a boolean

shravan97: Sure @etejedor . That sounds better 👍

PyRDF/backend/Backend.py
etejedor: I think `supported_operations` in `Backend` should contain absolutely everything, including what is now in `future_operations`. Then the rest of the backends can redefine those containers.

shravan97: @etejedor , I think it'd better if `Operation` class' `operations_dict` itself contains all the operations including the future ones. Because, the ones in `Operation` is immaterial of the backend right ?

shravan97: But @etejedor , what would be the operation type of Instant actions like `Snapshot` or `Foreach` ?

etejedor: For me they are both actions.

etejedor: @diocas @JavierCVilla @bluehood opinions here?

shravan97: Ok @etejedor . I'll keep them as actions for now 👍

shravan97:
@etejedor , made those changes now 👍

PyRDF/backend/Local.py
```diff
@@ -13,6 +13,41 @@ class Local(Backend):
...
+        Parameters
+        ----------
+        config (optional)
+            The config object for the required
+            backend. The default value is an
+            empty Python dictionary `{}`.
+
+        """
+        super().__init__(config)
+        self.supported_operations = [
```
etejedor: I would expect the `Local` backend to copy the list of the superclass and remove those items that are not supported.

PyRDF/backend/Local.py
```diff
@@ -13,6 +13,41 @@ class Local(Backend):
...
+        'Profile3D',
+        'Count',
+        'Min',
+        'Max',
+        'Mean',
+        'Sum',
+        'Fill',
+        'Report',
+        'Range',
+        'Take',
```
etejedor: `Take` is not supported by the local backend, and I believe `Snapshot` and `Foreach` are not supported either?

PyRDF/backend/Local.py
```diff
@@ -13,6 +13,41 @@ class Local(Backend):
...
+        'Profile2D',
+        'Profile3D',
+        'Count',
+        'Min',
+        'Max',
+        'Mean',
+        'Sum',
+        'Fill',
+        'Report',
+        'Range',
```
etejedor: We can leave `Range` in the list only if the `Local` backend is not configured to use multi-threading.

PyRDF/backend/Local.py
shravan97: Oh yes @etejedor . Forgot to remove those. Shall do that 👍

PyRDF/backend/Local.py
shravan97: Sure @etejedor . Will do this 👍

shravan97:
@etejedor , made that change now 👍

PyRDF/backend/Local.py
etejedor: Ping :D

shravan97: @etejedor , done now :+1:

PyRDF/backend/Local.py
```diff
@@ -13,6 +13,26 @@ class Local(Backend):
...
+            empty Python dictionary `{}`.
+
+        """
+        super().__init__(config)
+        operations_not_supported = [
+        'Take',
+        'Snapshot',
+        'Foreach'
+        ]
+        self.supported_operations = [op for op in self.supported_operations if op not in operations_not_supported]
```
etejedor:
```python
if not ROOT.ROOT.IsImplicitMTEnabled():
  self.supported_operations.append('Range')
```

shravan97: @etejedor , can I give this as a part of the multi threading PR ?

tests/test_backend.py
```diff
@@ -67,3 +67,41 @@ class TestBackend(PyRDF.backend.Backend.Backend):
 
         with self.assertRaises(TypeError):
             TestBackend()
+
+class OperationSupportTest(unittest.TestCase):
+    """
+    Test cases to ensure that incoming operations are
+    classified accurately in local environment.
+
+    """
```
etejedor, line 75: We need to test all the operations that are not supported by every backend.
In the case of local, we need to test that Take, Snapshot and Foreach raise an exception. Also, we need to test that Range only raises an exception with the local backend in multi-threaded mode, but not in sequential mode.

PyRDF/backend/Local.py
shravan97: And also, I can use `config` instead ROOT itself right ?

etejedor: I suggest you use for now the IsImplicitEnabled method, since we haven't yet introduced the config options for the local backend. Anyway, you can always rely on that method, even if you check the config option before and activate the implicit multi-threading yourself in the Local constructor.

Let's just include this line in this PR already, it does not hurt.

shravan97: Ok @etejedor :+1:

tests/test_backend.py
shravan97: Ok @etejedor . Will add them :+1:

PyRDF/backend/Local.py
shravan97: There's one more issue with this @etejedor . What if the user enables multi-threading after the creation of this constructor ?

shravan97: So, in that case, I think I must add the check to `check_supported` itself then

shravan97: I've made that change anyways @etejedor 👍

etejedor: Yes it's a bit tricky, even adding a check in `check_supported` would not completely solve the issue, because the user might activate this flag after all the nodes of the graph have been created and right before the actual execution (access to a Proxy).
I think for now we can just assume that IMT will be activated already before the Local backend is constructed.

etejedor (approved):
LGTM, I believe we can merge this

PyRDF/backend/Local.py
shravan97: Yes @etejedor . I agree. We can look into more cases in the multi-threading PR, which is coming up next 👍

shravan97:
Ok @etejedor . I'll rebase and notify 👍

PyRDF/backend/Backend.py
```diff
@@ -30,6 +51,25 @@ def __init__(self, config={}):
         """
         self.config = config
 
+    def check_supported(self, operation_name):
+        """
+        Checks if a given operation is supported
+        in Local backend.
```
JavierCVilla: @shravan97 why "in Local" here? Isn't this the abstract class?

shravan97: @JavierCVilla  thanks for pointing out. Shall I make this a part of a new PR ?

JavierCVilla: I would just add it as part of this PR.

shravan97: Ok @JavierCVilla . Will fix 👍

shravan97:
@etejedor , rebased now 👍

JavierCVilla (approved):

PR #32, opened 2018-06-26
Add tutorials with multi-threading enabled 

shravan97:
Features in this PR : 
* Re-structure the tutorials directory into : 
 ```
tutorials/
     |
      ----- local
                   |
                    ------sequential/.......
                    ------MT/.......
```
* Enable multi-threading for all tutorials and add them to `tutorials/local/MT`

### CODE REVIEW

File: tutorials/local/MT/df001_introduction.py [hunk 1]
```diff
@@ -0,0 +1,111 @@
...
+fileName = "df001_introduction_py.root"
+treeName = "myTree"
+fill_tree(treeName, fileName)
+
+
+# We read the tree from the file and create a RDataFrame, a class that
+# allows us to interact with the data contained in the tree.
+RDF = PyRDF.RDataFrame
+d = RDF(treeName, fileName)
+ROOT.ROOT.EnableImplicitMT()
```
etejedor: If you enable it here, it won't run multi-threaded because the local backend has already been constructed, right?
You should start all the tests with
```python
import ROOT
ROOT.ROOT.EnableImplicitMT()
import PyRDF
```

shravan97: Ok @etejedor 👍

shravan97: But @etejedor , In the current code, everything apart from building the root file using snapshot, everything else will run multi-threaded right ?
Because, I'm not enabling multi-threaded anywhere in PyRDF right ?

shravan97: @etejedor , I've made that change now anyway 👍

etejedor: Right, it will run MT but we still have to enable it before constructing the local backend, just in case there is a Range that we need to detect.

shravan97:
@etejedor , this directory structure means, I can add [this file](https://github.com/root-project/root/blob/master/tutorials/dataframe/df006_ranges.py) into `local/sequential`. Shall I push a new commit now for this ?

etejedor:
@shravan97 can you adapt the script to run the tests to also run the MT tutorials?

shravan97:
@etejedor . Oh yes. Thanks for reminding me. Will do that 👍

shravan97:
@etejedor , done 👍

PR #33, opened 2018-06-26
Add more operations to operations_dict

shravan97:
Features in this PR : 
* Add more operations to `operations_dict`
* Add one more operation type (`INSTANT_ACTION`) correctly classify operations like `Snapshot` and `Foreach`
* Add a unit test to check classification of instant actions

### DIFF
--- a/PyRDF/Operation.py
+++ b/PyRDF/Operation.py
@@ -59,7 +59,7 @@ class Operation(object):
 
     """
 
-    Types = Enum("Types", "ACTION TRANSFORMATION")
+    Types = Enum("Types", "ACTION TRANSFORMATION INSTANT_ACTION")
 
     def __init__(self, name, *args, **kwargs):
         """
@@ -94,6 +94,8 @@ class Operation(object):
         operations_dict = {
         'Define':ops.TRANSFORMATION,
         'Filter':ops.TRANSFORMATION,
+        'Range':ops.TRANSFORMATION,
+        'Aggregate':ops.ACTION,
         'Histo1D':ops.ACTION,
         'Histo2D':ops.ACTION,
         'Histo3D':ops.ACTION,
@@ -106,9 +108,13 @@ class Operation(object):
         'Mean':ops.ACTION,
         'Sum':ops.ACTION,
         'Fill':ops.ACTION,
-        'Report':ops.ACTION
+        'Reduce':ops.ACTION,
+        'Report':ops.ACTION,
+        'Take':ops.ACTION,
+        'Snapshot':ops.INSTANT_ACTION,
+        'Foreach':ops.INSTANT_ACTION
         }
-        
+
         op_type = operations_dict.get(name)
 
         if not op_type:
--- a/PyRDF/backend/Backend.py
+++ b/PyRDF/backend/Backend.py
@@ -70,7 +70,7 @@ class Backend(ABC):
 
         """
         if operation_name not in self.supported_operations:
-            raise Exception("Invalid operation \"{}\"".format(operation_name))
+            raise Exception("The current backend doesn't support \"{}\" !".format(operation_name))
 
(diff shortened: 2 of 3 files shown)


shravan97:
@etejedor , I'd forgotten earlier to add more operation to `operations_dict`. So, I took this PR to add those and also was forced to add one more operation type (INSTANT_ACTION) due to `Snapshot` and `Foreach`

diocas:
Hi Shravan,

I haven't been paying much attention to these last PR's, so please tell me if I'm assuming or discussing something that was already discussed before.

One question about this PR in particular:
* In `is_action()` (Operation.py#L123), you only check if the type is `ACTION`. Isn't an `INSTANT_ACTION` also an action? Or you will then write a `is_instant_action` method?
    * In reality, you're not using these instant actions for anything, because they're not even supported in any backend... Or are they going to be supported in Dist?
    * And, for me, given that the error message is always the same (see bellow), I don't even see the point of having these unsupported operations here...

Other questions:
* The `Take` action is not part of the `operations_dict ` (Operation.py#L94), even though it appears in the Backend.py  `supported_operations`. Shouldn't it be?
* There are 4 operations that are not supported at all, independently of the backend, correct? So, wouldn't it be better if they raise an exception in Operation (as we had before, I think)?

About this last question, did @etejedor agreed that we should remove this exception from the Operation itself?
Because, maybe it would be better to throw an exception when you try to add "not supported at all" operations. And something more explicit than the current "Invalid Operation" exception, that you raise in Operation and Backend. As I said before, if some operations are not supported in any Backend, and the error message is like the one it raises when the user wrote something not recognizable, why do we keep these operation in the supported list?

We could keep them there, but we should have different kinds of exceptions, like these:
 * Invalid operation: the operation the user tried to use does not exist:
 * Not supported operation: this operation that the user possibly knows from RDF documentation is not supported here;
 * Not supported operation in this backend: this operation is supported, but not for this backend.

Diogo

diocas:
Also, is it good idea (was this already discussed?) to remove not supported operations from the `supported_operations` list? Shouldn't the backend *add* supported ops instead?

Imagine I create a new backend and some operations are not supported. Then, (py)RDF adds a new operation that might not be supported in my backend. If I was the one controlling what could be used, then this update wouldn't break anything on my backend by default. The way it's now, it breaks things by default, until I disable the option.

shravan97:
* @diocas , the idea was to make `operations_dict` a super set of all operations. This is because, the `Backend` only checks for support (according to the backend that was chosen) and then Operation classifies it. And since I added `Snapshot` onto `supported_operations` in the previous commit, I'm forced to add it here too. And since `Snapshot` is not an action, I had to add it as an `INSTANT_ACTION`. But as you said, every instant action is an action. Hence it's safe enough to remove `INSTANT_ACTION` enum type for now. 

* I missed out the `Take` operation here by mistake. I'll surely add it and push 
* I also agree that the error message too generic in `Backend.py`. I'll definitely have to change that 
* And I feel it's better to raise the support error in `Backend` itself as it's the one checking for support. What we're doing here is we're calling `supported_operations` function from inside `Node` class (L93). So, in that way, I feel it'd be better to keep the error inside `Backend`

shravan97:
@diocas and @etejedor , I've add `Take` to the `operations_dict` and changed the error message in `check_supported`. Is there anything else that I should change in this PR ? And also, is the error message fine ?

diocas:
It looks ok. I will aprove it.
Btw, just a reminder: like we discussed last week, you shouldn't force replace commits. We agreed that you will just submit new commits and then, **after it gets approved**, you squash them.

etejedor (approved):

shravan97:
Yup @diocas . I agree. Did a force push here because, the changes were too small. Will add more commits on top from next PR 👍

PR #37, opened 2018-07-03
Add a function to compute number of entries in the input dataset

shravan97:
Features in this PR : 
* Implement `get_num_entries` in `RDataFrame` to get the number of entries in the input dataset
* Add unit tests for `get_num_entries()`  method

### DIFF
--- a/PyRDF/RDataFrame.py
+++ b/PyRDF/RDataFrame.py
@@ -78,6 +78,37 @@ class RDataFrame(Node):
 
         return reqd_vec
 
+    def get_num_entries(self):
+        """
+        Gets the number of entries in
+        the given dataset.
+
+        """
+        first_arg = self.args[0]
+        if isinstance(first_arg, int):
+            # If there's only one argument
+            # which is an integer, return it.
+            return first_arg
+        elif isinstance(first_arg, ROOT.TTree):
+            # If the argument is a TTree or TChain,
+            # get the number of entries from it.
+            return first_arg.GetEntries()
+
+        second_arg = self.args[1]
+
+        # Construct a ROOT.TChain object
+        chain = ROOT.TChain(first_arg)
+
+        if isinstance(second_arg, str):
+            # If the second argument is a string
+            chain.Add(second_arg)
+        else:
+            # If the second argument is a list or vector
+            for fname in second_arg:
+                chain.Add(fname)
+
+        return chain.GetEntries()
+
 
 class RDataFrameException(Exception):
     """
--- a/.travis-script.sh
+++ b/.travis-script.sh
@@ -3,5 +3,6 @@ sudo apt-get --yes install python-pip
 sudo pip install enum34
 sudo pip install nose
 source /usr/local/bin/thisroot.sh
+sudo chown -R $UID:$UID /app
 cd app
 nosetests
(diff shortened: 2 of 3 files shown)


PyRDF/RDataFrame.py
```diff
@@ -78,6 +78,31 @@ def _get_vector_from_list(self, arg):
...
+        """
+        Gets the number of entries in
+        the given dataset.
+
+        """
+        if isinstance(self.args[0], int):
+            # If there's only one argument
+            # which is an integer, return it.
+            return self.args[0]
+
```
etejedor: Can we add here:
```python
elif isinstance(self.args[0], ROOT.TTree):
    return self.args[0].GetEntries()
```
This is covering this constructor:
https://root.cern/doc/master/classROOT_1_1RDataFrame.html#a6a45ce7601ffbca837f6773cd55a3e72

etejedor: Since they are repeated several times, can we have local variables for:
```python
first_arg = self.args[0]
sec_arg = self.args[1]
```
**right before** the first time they are used?

PyRDF/RDataFrame.py
```diff
@@ -78,6 +78,31 @@ def _get_vector_from_list(self, arg):
...
+        # Construct a ROOT.TChain object
+        import ROOT
```
etejedor: We do not need this import here, it is imported at the beginning of this module.

PyRDF/RDataFrame.py
shravan97: Sounds good @etejedor . I'll make both these changes 👍

shravan97:
Fixed those issues here @etejedor . I'll rebase the commits when you're okay with this PR 👍

PyRDF/RDataFrame.py
shravan97: Shall I also add a test for the `ROOT.TTree` case @etejedor ?

PyRDF/RDataFrame.py
```diff
@@ -84,21 +84,26 @@ def get_num_entries(self):
...
 
         """
-        if isinstance(self.args[0], int):
+        first_arg = self.args[0]
+        if isinstance(first_arg, int):
             # If there's only one argument
             # which is an integer, return it.
-            return self.args[0]
+            return first_arg
+
```
etejedor: No need of newline here

PyRDF/RDataFrame.py
```diff
@@ -84,21 +84,26 @@ def get_num_entries(self):
...
+        elif isinstance(first_arg, ROOT.TTree):
+            return first_arg.GetEntries()
```
etejedor: To be consistent with the other comments: I would add a comment before this line: "The argument is a TTree or TChain, get the number of entries from it"

PyRDF/RDataFrame.py
etejedor: Yes, please add another test.

shravan97: 👍

shravan97:
@etejedor I've added a test now. Anything else in this PR ?

shravan97:
@etejedor and @diocas , done rebasing this. Added 3 commits, one for `get_num_entries` function, one for tests and one for that change in travis script

PR #38, opened 2018-07-03
Change Callable implementation

shravan97:
Features in this PR : 
* Change the implementation of the callable function to return only the values of the executed action nodes
* Add a new function to get the list of action nodes in the same order as they were executed 
* Modify tests to incur the implementation changes

### DIFF
--- a/PyRDF/CallableGenerator.py
+++ b/PyRDF/CallableGenerator.py
@@ -21,6 +21,44 @@ class CallableGenerator(object):
         """
         self.head_node = head_node
 
+    def get_action_nodes(self, node_py=None):
+        """
+        Recurses through PyRDF graph and collects
+        the PyRDF nodes.
+
+        Parameters
+        ----------
+        node_py (optional)
+            The current state's PyRDF node. If `None`,
+            it takes the value of `self.head_node`.
+
+        Returns
+        -------
+            list
+                A list of the action nodes of the graph in DFS order, which
+                coincides with the order of execution in the callable function.
+
+        """
+        return_nodes = []
+
+        if not node_py:
+            # In the first recursive state, just set the
+            # current PyRDF node as the head node
+            node_py = self.head_node
+        else:
+            if node_py.operation.is_action():
+                # Collect all action nodes in order to return them
+                return_nodes.append(node_py)
+
+        for n in node_py.children:
+            # Recurse through children and collect them
+            prev_nodes = self.get_action_nodes(n)
+
+            # Attach the children nodes
+            return_nodes.extend(prev_nodes)
+
+        return return_nodes
+
     def get_callable(self):
         """
         Converts a given graph into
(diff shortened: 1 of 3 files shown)


PyRDF/backend/Local.py
```diff
@@ -51,7 +51,9 @@ def execute(self, generator):
 
         rdf = ROOT.ROOT.RDataFrame(*generator.head_node.args) # Create RDF object
 
-        values, nodes = mapper(rdf) # Execute the mapper function
+        values = mapper(rdf) # Execute the mapper function
+        nodes = generator.get_action_nodes()
+        # Get the action nodes in the same order as values
```
etejedor: Should this comment precede (or appear next to) the previous line?

PyRDF/CallableGenerator.py
```diff
@@ -21,6 +21,45 @@ def __init__(self, head_node):
...
+
+        """
+        return_nodes = []
+
+        if not node_py:
+            # In the first recursive state, just set the
+            # current PyRDF node as the head node
+            node_py = self.head_node
+        else:
+
```
etejedor: Remove extra newline?

PyRDF/backend/Local.py
shravan97: It should appear on line 56 after `nodes = generator.get_action_nodes()`. I put it below as that line would get too long in that case

PyRDF/CallableGenerator.py
shravan97: I'll remove this @etejedor 👍

PyRDF/CallableGenerator.py
```diff
@@ -21,6 +21,45 @@ def __init__(self, head_node):
...
+        ----------
+        node_py (optional)
+            The current state's PyRDF node. If `None`,
+            it takes the value of `self.head_node`.
+
+        Returns
+        -------
+            list
+                A list of action nodes in the same order as the
+                order of execution in the callable function.
```
etejedor: Can we say: A list of the action nodes of the graph in DFS order, which coincides with the order of execution in the callable function.

shravan97: Yup @etejedor . That sounds better. I'll modify it 👍

PyRDF/CallableGenerator.py
```diff
@@ -49,20 +88,21 @@ def mapper(node_cpp, node_py=None):
...
+
             Returns
             -------
-                tuple
-                    This tuple consists of 2 lists, the first one
-                    consisting of output values of action nodes and
-                    the second one of action nodes in the same order
-                    as that of the values list.
+                list
+                    This is a list of values in the order that they
```
etejedor: And here I would say: "A list of `TResultPtr` objects in DFS order of their corresponding actions in the graph."

@shravan97 , note that here we do not execute anything, the whole event loop will be triggered when accessing one of the `TResultPtr`s later in the backend.

etejedor:
Once the comments I made are applied, I think this is good to go. It is a pity that we have some code repetition because of the double traversal of the graph, but we need to do it this way because the distributed backends will only get the corresponding nodes at the very end, in the client. Having the values of the nodes in the workers is useless.

PyRDF/backend/Local.py
etejedor: Then I would put the comment in the line above.

PyRDF/CallableGenerator.py
shravan97: Yup @etejedor . Will change this 👍

PyRDF/backend/Local.py
shravan97: I'll make the change @etejedor  👍

shravan97:
Yes @etejedor . I agree. I believe there's just a difference of 1 line between `get_action_nodes` and `mapper`

shravan97:
@etejedor done now 👍

shravan97:
@etejedor and @diocas , rebase done 👍

PR #39, opened 2018-07-04
Make 'Node' objects serializable

shravan97:
Features in this PR : 
* Make objects of `Node` class serializable (picklable) by implementing `__getstate__` and `__setstate__` dunder methods
* Add unit tests to check `Node` serializability

### DIFF
--- a/PyRDF/Node.py
+++ b/PyRDF/Node.py
@@ -61,6 +61,38 @@ class Node(object):
         self._cur_attr = "" # Name of the new incoming operation
         self.value = None
 
+    def __getstate__(self):
+        """
+        Converts the state of the current node
+        to a Python dictionary.
+
+        Returns
+        -------
+        Python Dictionary
+            A dictionary that stores all instance variables
+            that represent the current PyRDF node.
+
+        """
+        state_dict = {'children':self.children}
+        if self.operation:
+            state_dict['operation_name'] = self.operation.name
+            state_dict['operation_args'] = self.operation.args
+            state_dict['operation_kwargs'] = self.operation.kwargs
+
+        return state_dict
+
+    def __setstate__(self, state):
+        """
+        Retrieves the state dictionary of the current
+        node and sets the instance variables.
+
+        """
+        self.children = state['children']
+        if state.get('operation_name'):
+            self.operation = Operation(state['operation_name'], *state['operation_args'], **state["operation_kwargs"])
+        else:
+            self.operation = None
+
     def __getattr__(self, attr):
         """
         Intercepts any call to the current node and dispatches
@@ -81,6 +113,13 @@ class Node(object):
         """
 
         self._cur_attr = attr # Stores new operation name
+
+        # Check if the current call is a dunder method call
+        import re
+        if re.search("^__[a-z]+__$", attr):
+            # Raise an AttributeError for all dunder method calls
+            raise AttributeError("Such an attribute is not set ! ")
+
         return self._call_handler
 
     def _call_handler(self, *args, **kwargs):
--- a/tests/test_node.py
+++ b/tests/test_node.py
@@ -265,3 +265,141 @@ class DfsTest(unittest.TestCase):
         reqd_order = [1, 3, 2, 2, 3, 2]
 
         self.assertEqual(obtained_order, reqd_order)
+
+class DunderMethodsTest(unittest.TestCase):
+    """
+    Test cases to check the response of
+    the Node class for various dunder
+    method calls.
+
+    """
+    def test_get_state(self):
+        """
+        Test cases to check the working of
+        __getstate__ method on Node class.
+
+        """
+        node = Node(None, None) # Head node
+        n1 = node.Define("a", b="c") # First child node
+
+        # Required dictionaries
+        node_dict = {"children":[n1]}
+        n1_dict = {
+        'operation_name':"Define",
+        'operation_args':("a",),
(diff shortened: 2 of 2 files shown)


etejedor:
For this one, I need to understand why a `Node` is not pickable as it is and we need these extra methods. This is something we can discuss tomorrow in the call.

eguiraud:
@etejedor @shravan97 
Enric and I brainstormed this a little bit more:
on one hand, it's highly possible that _now that `__getattr__` correctly raises an `AttributeError`__ instead of a `RDataFrame` exception, `__setstate__` and `__getstate__` are not required: pickle would probably fallback to some other mechanism seeing that `__getstate__` is not implemented.

On the other hand, `__getstate__` and `__setstate__` correctly implement and document the fact that we don't want to serialize the `get_head` lambda and its contents (which we want anyway).

So in the end the two methods can probably stay, as the pickle-friendly way to specify exactly what is the information that we want to be serialized.

shravan97:
@etejedor and @bluehood `get_head` is not serializable as it is a lambda. And it won't be serializable even if it is a function. This means, we could recursively disable the `get_head` lambdas from all the nodes. But then, I guess implementing a `__getstate__` and `__setstate__` is a better solution

eguiraud:
@shravan97 as we discussed last weekend, spark is (somehow) able to serialize callables, otherwise `map` wouldn't work.

shravan97:
Yup I agree @bluehood . But lambdas aren't serializable though. Check the 5th point in this [link](https://docs.python.org/3/library/pickle.html#what-can-be-pickled-and-unpickled) . And even if we change `get_head` to a function, there would be cyclic dependency right ?

eguiraud:
@shravan97 again: spark is smarter than pure pickle. for example, this works just fine:
```python
sc = SparkContext.getOrCreate()
vals = [1, 2, 3, 4]
r1 = sc.parallelize(vals, 4)
r1.map(lambda x : x*x).treeReduce(lambda x, y: x + y) # outputs 30
```

even if we _don't_ change `get_head` to a function, there might be dependency issues, and with your `get_state` we get two birds with one stone:
* we can be sure that no cycling references will be present at the serialization step
* we can be sure that only the content we care about is serialized

shravan97:
Nice examples and good explanation @bluehood

shravan97:
@etejedor and @diocas I'm done with the changes here and I've rebased it as well 👍

PyRDF/Node.py
```diff
@@ -61,6 +61,40 @@ def __init__(self, get_head, operation):
...
+
+        Returns
+        -------
+        Python Dictionary
+            A dictionary that stores all instance variables
+            that represent the current PyRDF node.
+
+        """
+        state_dict = {'children':self.children}
+        if not self.operation:
```
diocas: why not `if self.operation` and then you remove the return from here?

PyRDF/Node.py
```diff
@@ -61,6 +61,40 @@ def __init__(self, get_head, operation):
...
+            return state_dict
+
+        state_dict['operation_name'] = self.operation.name
```
diocas: I also would say that maybe you could do something like `state_dict.update(self.operation.__dict__)`, no?
It adds an extra field, but it would be cleaner.

diocas (changes requested):

PyRDF/Node.py
shravan97: @diocas , `operation.__dict__` would also have the Enum variable named `Types`. It is not needed while pickling right ?

PyRDF/Node.py
shravan97: The amount of code is same for both right ? 
```python
        state_dict = {'children':self.children}
        if self.operation:
            state_dict['operation_name'] = self.operation.name
            state_dict['operation_args'] = self.operation.args
            state_dict['operation_kwargs'] = self.operation.kwargs
            return state_dict

        return state_dict
```

shravan97: Is this what you meant @diocas ?

PyRDF/Node.py
diocas: That's why I said that it was *only* one extra field. But I forgot it was an Enum, so forget this suggestion. As mixing Enums and pickle is not a good idea..

PyRDF/Node.py
diocas: Yes :)

diocas: oh sorry, looked too quickly... It's not what I meant.. This is:
```python
        state_dict = {'children':self.children}
        if self.operation:
            state_dict['operation_name'] = self.operation.name
            state_dict['operation_args'] = self.operation.args
            state_dict['operation_kwargs'] = self.operation.kwargs

        return state_dict
```

shravan97: Ah ok. Right. Will make this change 👍

PyRDF/Node.py
shravan97: Is this fine @diocas ?
```python
dict = self.operation.__dict__
dict.pop('op_type')
state_dict.update(dict)
```

diocas: meh
I don't really like this one.. keep it as is

shravan97: Ok 👍

shravan97:
Any other change @diocas ? If not, I'll make this change and rebase directly

diocas:
No, go ahead

shravan97:
@diocas , done now 👍

PR #41, opened 2018-07-05
Add more details to README

shravan97:
Features in this PR : 
* Modify readme to add more details and sample code


diocas:
Can you please add some information about GSoC and HSF/CERN?
I think it should be mentioned you are doing this because of that.

shravan97:
Ok @diocas 👍

shravan97:
Is this fine @diocas ?

PR #45, opened 2018-07-10
Implement Spark backend and write tests

shravan97:
Features in this PR : 
* Add more member functions to `Dist` class and make it abstract
* Implement Spark backend
* Write tests for the above

### DIFF
--- a/PyRDF/backend/Dist.py
+++ b/PyRDF/backend/Dist.py
@@ -1,13 +1,199 @@
 from __future__ import print_function
 from .Backend import Backend
+from .Local import Local
+from abc import abstractmethod
 
 class Dist(Backend):
+    """
+    Base class for implementing all distributed backends.
+
+    """
+    def __init__(self, config = {}):
+        """
+        Creates an instance of Dist.
+
+        Parameters
+        ----------
+        config : dict (optional)
+            The config options for the current
+            distributed backend. Default value is
+            an empty python dictionary `{}`.
+
+        """
+        super(Dist, self).__init__(config)
+        # Operations that aren't supported in distributed backends
+        operations_not_supported = [
+        'Range',
+        'Take',
+        'Snapshot',
+        'Foreach',
+        'Reduce',
+        'Aggregate'
+        ]
+        self.supported_operations = [op for op in self.supported_operations if op not in operations_not_supported]
+
+    def BuildRanges(self, nentries, npartitions):
+        """
+        Builds range pairs from the given values
+        of the number of entries in the dataset and
+        number of partitions required.
+
+        Parameters
+        ----------
+        nentries : int
+            The number of entries in a dataset.
+
+        npartitions : int
+            The number of parallel computations.
+
+        Returns
+        -------
+        Python list
+            This is a list of range pairs (represented here
+            as 2-member tuples).
+
+        """
+        if npartitions > nentries:
+            # Restrict 'npartitions' if it's greater
+            # than 'nentries'
+            npartitions = nentries
+
+        partition_size = int(nentries/npartitions)
+
+        i = 0 # Iterator
+
+        ranges = []
+
+        remainder = nentries%npartitions
+
+        while i < nentries:
+            # Start value of current range
+            start = i
+            end = i = start + partition_size
+
+            if remainder:
+                # If the modulo value is not
+                # exhausted, add '1' to the end
+                # of the current range
+                end = i = end+1
+                remainder -= 1
+
+            ranges.append((start, end))
+
+        return ranges
 
     def execute(self, generator):
         """
-        Execution of the event-loop
-        in distributed environment
+        Executes the current RDataFrame graph
+        in the given distributed environment.
+
+        Parameters
+        ----------
+        generator : PyRDF.CallableGenerator.CallableGenerator
+            An instance of type `CallableGenerator` that is
+            responsible for generating the callable function.
 
         """
-        print("Distributed execution "\
-              "hasn't been implemented yet !")
+        callable_function = generator.get_callable()
+        # Arguments needed to create PyROOT RDF object
+        rdf_args = generator.head_node.args
+
+        def mapper(current_range):
+            """
+            Triggers the event-loop and executes all
+            nodes in the computational graph using the
+            callable.
+
+            Parameters
+            ----------
+            current_range : tuple
+                A pair that contains the starting and ending
+                values of the current range.
+
+            Returns
+            -------
+            Python list
+                This respresents the list of values of all
+                action nodes in the computational graph.
+
+            """
+            import ROOT
+            rdf = ROOT.ROOT.RDataFrame(*rdf_args) # PyROOT RDF object
+
+            # TODO : If we want to run multi-threaded in a Spark node in
+            # the future, use `TEntryList` instead of `Range`
+            rdf_range = rdf.Range(current_range[0], current_range[1])
+
+            output = callable_function(rdf_range) # output of the callable
+
+            for i in range(len(output)):
+                # FIX ME : RResultPtrs aren't serializable,
+                # because of which we have to manually find
+                # out the types here and copy construct the
+                # values.
+
+                # The type of the value of the action node
+                value_type = type(output[i].GetValue())
+                # The `value_type` is required here because,
+                # after a call to `GetValue`, the values die
+                # along with the RResultPtrs
+                output[i] = value_type(output[i].GetValue())
+
+            return output
+
+        def reducer(values_list1, values_list2):
+            """
+            Merges two given lists of values that were
+            returned by the mapper function for two different
+            ranges.
+
+            Parameters
+            ----------
+            values_list1
+                A list of computed values for a given entry
+                range in a dataset.
+
+            values_list2
+                A list of computed values for a given entry
+                range in a dataset.
+
+            Returns
+            -------
+            Python list
+                This is a list of values obtained after
+                merging two given lists.
+
+            """
+            import ROOT
+            for i in range(len(values_list1)):
+                # A bunch of if-else conditions to merge two values
+                if isinstance(values_list1[i], ROOT.TH1) or isinstance(values_list1[i], ROOT.TH2):
+                    # Merging two objects of type ROOT.TH1D or ROOT.TH2D
+                    values_list1[i].Add(values_list2[i])
+                elif isinstance(values_list1[i], int):
+                    # Merging two objects of type `int`
+                    values_list1[i] += values_list2[i]
+
+            return values_list1
+
+        # Get number of entries in the input dataset using
+        # arguments passed to RDataFrame constructor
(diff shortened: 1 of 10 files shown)


shravan97:
@ mentors, this is a massive PR. I guess it'd be better to discuss this during our meeting. Besides, if any of you have suggestions on splitting the commits further or splitting into different PRs itself, please do let me know :D

shravan97:
@etejedor and mentors , I've made the changes that we'd discussed during the last meeting 👍

PyRDF/backend/Spark.py
```diff
@@ -0,0 +1,67 @@
...
+        super(Spark, self).__init__(config)
+
+        try:
+            # Remove the value of 'npartitions' from config dict
+            self.npartitions = config.pop('npartitions')
+        except KeyError:
+            # Set the default value to 2 if 'npartitions'
+            # doesn't exist
+            self.npartitions = 2
+
```
shravan97: @etejedor and mentors, I feel its not very clean to pop out only `npartitions` from `config` and set the rest as `SparkConf`. What do you guys think ?

shravan97:
@etejedor , the integration test for Spark is also ready. But for that, I have to make changes in `.travis-script.sh` and `run_tests.sh` too. So, I'll give that in a separate PR 👍

PyRDF/backend/Spark.py
diocas: Why?
Other thing, what was discussed was that if no `npartitions` was provided, you would take the value of number of executors, and not a random 2.

PyRDF/backend/Spark.py
```diff
@@ -0,0 +1,67 @@
...
+        sparkConf = SparkConf().setAll(config.items())
+        self.sparkContext = SparkContext.getOrCreate(sparkConf)
```
diocas: If a SparkContext exists (like the case of SWAN), these configs are ignored. I guess there's no problem, as the user was the one that created the config. Do you imagine any problem with this?
Should we check if there's already a sc and give a warning or it's not really necessary?

tests/unit/test_backend_dist.py
```diff
@@ -0,0 +1,275 @@
...
+        method throws a `TypeError`.
+
+        """
+        class TestBackend(backend.Dist):
+            pass
+
+        with self.assertRaises(TypeError):
+            TestBackend()
+
+class SparkBackendInitTest(unittest.TestCase):
```
diocas: I'm not understanding the way you divided the tests. Didn't we agree that you should have separate files for Spark, Dist and Common? I'm seeing Spark and Dist together...

diocas: And if this is Spark related, it should have spark in the filename.

diocas (changes requested):
From what I took note, I think you address the majority of issues mentioned (no zero division ex, fall back to local when 0 entries, new range in mapper, some separation of tests, etc). There are just small things missing, so please address them (or disagree with me).
Then let's see if Enric has anything else to say.

PyRDF/backend/Spark.py
shravan97: @diocas , me and @etejedor agreed that the value has to be `2` from `spark.executor.instances` in this link : [https://spark.apache.org/docs/latest/running-on-yarn.html#spark-properties](https://spark.apache.org/docs/latest/running-on-yarn.html#spark-properties) . And the `2` is compulsorily required here because, we have to create the range pairs only based on that

PyRDF/backend/Spark.py
shravan97: @diocas , yes. That'd be a useful message. It'd be great to show a warning. Let's wait and see what @etejedor has to say about this one

tests/unit/test_backend_dist.py
shravan97: @diocas , I think we decided to split them by `dist`, `local` and `common`.  Am I right @etejedor ?

PyRDF/backend/Spark.py
diocas: If you agreed, it's ok.

But my point was (and what I took note from the meeting):
 * the user has `npartition`? use it.
 * Otherwise use the number of executors from SparkConf
    * This value does not exist? therefore is 2 (the default one)
    * This value exists? then it can be used

Here you pick a random 2, that might not be the real value of executors. Remember that ` SparkContext.getOrCreate` can pick an already existing `SparkContext` that contains a different `SparkConf`...

shravan97: @diocas , That's a good point. How about just taking the number of partitions as `spark.executor.instances` in the `config` dict ? That is how `SparkConf` stores it

shravan97: And this way, I don't have to pop it out of the `config` dict as well. I can directly pass it on to `SparkConf`

shravan97: @diocas , addressed your issue now. I misunderstood it earlier, but fixed it now. Thanks for pointing out :)

PyRDF/backend/Spark.py
```diff
@@ -0,0 +1,67 @@
...
+        """
+        super(Spark, self).__init__(config)
+
+        try:
+            # Remove the value of 'npartitions' from config dict
+            self.npartitions = config.pop('npartitions')
+        except KeyError:
+            # Set the default value to 2 if 'npartitions'
+            # doesn't exist
+            self.npartitions = config.get('spark.executor.instances') or 2
```
diocas: Again, if a previous `SparkContext` is loaded, this value might not be the correct one...
Can't you just move this after line 40 and *then* you try to get the value from the `SparkConf` set in the `SparkContext`?

shravan97: Done now @diocas 👍

tests/unit/test_backend_dist.py
etejedor: Spark should also have its own file. Proposal:
- `PyRDF/tests/backend`
- `PyRDF/tests/backend/test_common.py`
- `PyRDF/tests/backend/test_local.py`  (if we need it)
- `PyRDF/tests/backend/test_dist.py`  (if we need it)
- `PyRDF/tests/backend/test_spark.py`

shravan97: Will fix this @etejedor :+1:

PyRDF/backend/Spark.py
etejedor: Yes, I agree we should show a warning if a `SparkContext` already exists and the user is trying to define new configurations.

PyRDF/backend/Spark.py
etejedor: I agree with @diocas that you should try to set the npartitions to the number of executors of the SparkContext. The logic you have now is good.

PyRDF/backend/Spark.py
```diff
@@ -0,0 +1,73 @@
...
+        except KeyError:
+            # Set the value to 'None' if 'npartitions'
+            # doesn't exist
+            self.npartitions = None
+
+        sparkConf = SparkConf().setAll(config.items())
+        self.sparkContext = SparkContext.getOrCreate(sparkConf)
+
+        # Set the value of 'npartitions' if it doesn't exist
+        self.npartitions = self.npartitions or self.sparkContext.getConf().get('spark.executor.instances') or 2
```
etejedor, line 49: Indeed this is the right logic:
- First try with user config for npartitions
- Second try to get default/user configured from SparkContext
- Last option, just give it a value (2 is fine).

PyRDF/backend/Spark.py
```diff
@@ -0,0 +1,73 @@
...
+            specifically reads three options, 'SparkContext'
+            or 'SparkConf' and 'npartitions'. If 'SparkContext'
+            is not given, 'SparkContext' uses 'SparkConf' option,
+            which by default is 'None' and 'npartitions' is
+            '2' by default.
+
+        """
+        super(Spark, self).__init__(config)
+
+        try:
```
etejedor: This `try` `except` can be replaced by just one line: `self.npartitions = config.pop('npartitions', None)`

shravan97: Oh yes. Thanks @etejedor . Will fix this 👍

PyRDF/backend/Dist.py
```diff
@@ -1,13 +1,199 @@
...
+            # than 'nentries'
+            npartitions = nentries
+
+        partition_size = int(nentries/npartitions)
+
+        i = 0 # Iterator
+
+        ranges = []
+
+        remaining = nentries%npartitions
```
etejedor: `remainder`?

tests/unit/test_backend_dist.py
```diff
@@ -126,6 +126,19 @@ def test_npartitions_with_num_executors(self):
...
 
+    def test_npartitions_with_already_existing_spark_context(self):
+        """
+        Test case to check that the value of `npartitions` is correctly set
+        when a Spark Context already exists.
+
+        """
+        from pyspark import SparkContext, SparkConf
+        sparkConf = SparkConf().set('spark.executor.instances', 15)
+        sc = SparkContext.getOrCreate(sparkConf)
```
etejedor: I would do here `sc = SparkContext(sparkConf)` instead of `getOrCreate`, in this test you want to make sure you create it with the configuration you are passing.

shravan97: 👍

etejedor:
The PR looks good to me modulo my last comments. Once implemented, we can merge.

shravan97:
Sure @etejedor . Will fix those. Are 4 commits okay for this PR ? 
* One commit for Dist
* One commit for Spark
* One for tests
* One for the scripts used for running the tests

etejedor:
@shravan97 sounds good.

shravan97:
@etejedor , I've fixed all comments now and rebased too 👍

diocas:
It's still missing the warning when a `SparkContext` shadows a configuration (Spark.py#46).

shravan97:
@diocas , me and @etejedor decided to just add a comment about the existence of a `SparkContext`. I've added an elaborate note in the Doc string of the `Dist` constructor here : [https://github.com/shravan97/PyRDF/pull/45/files#diff-4c8dc7bf86db5db2726e48aa5d70ff93R34](https://github.com/shravan97/PyRDF/pull/45/files#diff-4c8dc7bf86db5db2726e48aa5d70ff93R34)

diocas (approved):

etejedor:
@diocas the reason for what @shravan97 commented above is:
- There is no way to check in Spark if a context already exists (other than using an internal attribute of SparkContext), and thus to throw the warning we would need to compare the two configurations (user-provided and the one obtained from the context) after the context is getOrCreated.
- The current behaviour is consistent with the behaviour of `getOrCreate`: if the context already exists, the `conf` parameter is ignored.

diocas:
No problem, I just went through all commentaries to see what was missing and saw your last comment. I'll merge this PR now.

PR #46, opened 2018-07-18
Integration test for Spark backend

shravan97:
Features in this PR : 
* Convert the [GSoC tasks code](https://github.com/etejedor/GSoC_2018/blob/master/Spark_DistROOT_TDataFrame.ipynb) into an integration test and add it under Spark integration tests

### DIFF
--- a/.travis-script.sh
+++ b/.travis-script.sh
@@ -29,6 +29,7 @@ sudo python setup.py install
 nosetests tests/unit/*.py || exit 1
 nosetests tests/unit/backend/*.py || exit 1
 nosetests tests/integration/local/*.py || exit 1
+nosetests tests/integration/spark/*.py || exit 1
 
 # Run tutorials
 echo " ======== Running single-threaded tutorials ======== "
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -60,6 +60,11 @@ $nose_bin tests/integration/local/*.py || {
 	exit 1
 }
 
+$nose_bin tests/integration/spark/*.py || {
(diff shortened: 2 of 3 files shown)


tests/integration/spark/test_histograms.py
```diff
@@ -0,0 +1,72 @@
...
+        """
+        Cleans up the `SparkContext` objects
+        that were created.
+
+        """
+        from pyspark import SparkContext
+        context = SparkContext.getOrCreate()
+        context.stop()
+
+    def get_pyrdf_graph(self):
```
etejedor: Maybe `build_pyrdf_graph` instead of get? The rest looks good!

shravan97: Ok @etejedor . Will fix this. Shall I also do the final rebase after the fix ?

tests/integration/spark/test_histograms.py
```diff
@@ -0,0 +1,72 @@
+import unittest, PyRDF
+
```
etejedor: Perhaps we can call this file: `test_inv_mass.py`?

shravan97: Ok. Sounds good. I'll change this too 👍

tests/integration/spark/test_histograms.py
etejedor: Yes, you can keep the two commits you have now.

shravan97:
@etejedor , done resolving your comments 👍

PR #48, opened 2018-07-23
A new feature that enables the users to include headers by passing in a list

shravan97:
In this PR : 
* Create a static method in `Backend` class that declares required headers
* Create a new method `include` in `__init__.py` to accept headers that need to be declared
* Declare headers in `Local` and `Dist` classes before execution
* Write tests for all above features

### DIFF
--- a/PyRDF/__init__.py
+++ b/PyRDF/__init__.py
@@ -3,6 +3,7 @@ from .CallableGenerator import CallableGenerator
 from .backend import Local, Spark
 
 current_backend = Local()
+includes = []
 
 def use(backend_name, conf = {}):
 
@@ -20,3 +21,23 @@ def use(backend_name, conf = {}):
         current_backend = Spark(conf)
     else:
         raise Exception(" Incorrect backend environment \"{}\"".format(backend))
+
+def include(includes_list):
+    """
+    Includes a list of C++ headers to be
+    declared before execution.
+
+    parameters
+    ----------
+    includes_list : list or str
+        This list should consist of all necessary C++
+        headers as strings.
+
+    """
+    global current_backend, includes
+
+    if isinstance(includes_list, str):
+        # Convert to list if this is a string
+        includes_list = [includes_list]
+
+    includes.extend(includes_list)
--- a/PyRDF/backend/Dist.py
+++ b/PyRDF/backend/Dist.py
@@ -1,6 +1,7 @@
 from __future__ import print_function
 from .Backend import Backend
 from .Local import Local
+from .Utils import Utils
 from abc import abstractmethod
 
 class Dist(Backend):
@@ -98,6 +99,8 @@ class Dist(Backend):
         # Arguments needed to create PyROOT RDF object
         rdf_args = generator.head_node.args
 
+        from .. import includes
+
         def mapper(current_range):
             """
             Triggers the event-loop and executes all
@@ -118,6 +121,8 @@ class Dist(Backend):
 
             """
             import ROOT
+
+            Utils.declare_headers(includes) # Declare headers if any
             rdf = ROOT.ROOT.RDataFrame(*rdf_args) # PyROOT RDF object
 
             # TODO : If we want to run multi-threaded in a Spark node in
--- a/PyRDF/backend/Utils.py
+++ b/PyRDF/backend/Utils.py
@@ -1,0 +1,23 @@
+import ROOT
+
+class Utils(object):
+    """
+    Class that houses general utility
+    functions.
+
+    """
+    @classmethod
+    def declare_headers(cls, includes):
+        """
+        Declares all required headers using
+        PyROOT's "ROOT.gInterpreter.Declare".
+
+        parameters
+        ----------
+        includes : list
+            This list should consist of all necessary C++
+            headers as strings.
+
+        """
+        for header in includes:
+            ROOT.gInterpreter.Declare("#include \"{}\"\n".format(header))
--- a/tests/integration/local/test_headers_include.py
+++ b/tests/integration/local/test_headers_include.py
@@ -1,0 +1,26 @@
+import unittest, PyRDF, ROOT
+
+class IncludesLocalTest(unittest.TestCase):
+    """
+    An integration test to check that the required
+    header files are properly included.
+
+    """
+    def test_includes_function_with_filter_op(self):
+        """
+        An integration test to check that the filter
+        operation is able to use C++ functions that
+        were include using header files.
+
+        """
+        PyRDF.include("tests/integration/local/test_headers/header1.hxx")
+
+        rdf = PyRDF.RDataFrame(10)
+
+        # This filters out all numbers less than 5
+        rdf_filtered = rdf.Filter("check_number_less_than_5(tdfentry_)")
+        count = rdf_filtered.Count()
+
+        # The final answer should be the number of integers
+        # less than 5, which is 5.
+        self.assertEqual(count.GetValue(), 5)
--- a/tests/unit/backend/test_common.py
+++ b/tests/unit/backend/test_common.py
@@ -1,4 +1,5 @@
 import PyRDF, unittest, ROOT
+from PyRDF.backend.Utils import Utils
 
 class SelectionTest(unittest.TestCase):
     """
@@ -45,3 +46,87 @@ class BackendInitTest(unittest.TestCase):
 
         with self.assertRaises(TypeError):
             TestBackend()
+
+class IncludeHeadersTest(unittest.TestCase):
+    """
+    Tests to check the working of 'PyRDF.include' function.
+
+    """
+    def tearDown(self):
+        """
+        Resets the module-level variable
+        'includes' to an empty list.
+
+        """
+        PyRDF.includes = [] # reset includes
+
+    def test_default_empty_list_include(self):
+        """
+        Test case to ensure that 'PyRDF.include' function
+        raises a TypeError if no parameter is given.
+
+        """
+        with self.assertRaises(TypeError):
+            PyRDF.include()
+
+    def test_string_include(self):
+        """
+        Test case to check the working of 'PyRDF.include'
+        function when a single string is passed to it.
+
+        """
+        PyRDF.include("header1")
+
+        self.assertListEqual(PyRDF.includes, ["header1"])
+
+    def test_list_include(self):
+        """
+        Test case to check the working of 'PyRDF.include'
+        function when a list of strings is passed to it.
+
+        """
+        PyRDF.include(["header1"])
+
+        self.assertListEqual(PyRDF.includes, ["header1"])
+
+    def test_list_extend_include(self):
+        """
+        Test case to check the working of 'PyRDF.include'
+        function when different lists of strings are passed
+        to it multiple times.
+
+        """
+        PyRDF.include(["header1", "header2"])
+        PyRDF.include(["header3", "header4", "header5"])
+
+        self.assertListEqual(PyRDF.includes, ["header1", "header2", "header3", "header4", "header5"])
+
+class DeclareHeadersTest(unittest.TestCase):
+    """
+    Tests to check the working of the static
+    method 'declare_headers' in Backend class.
+
+    """
+    def test_single_header_declare(self):
+        """
+        Test case to check the working of 'declare_headers'
+        function when a single header needs to be included.
+
+        """
+        Utils.declare_headers(["tests/unit/backend/test_headers/header1.hxx"])
+
+        self.assertEqual(ROOT.f(1), True)
+
+    def test_multiple_headers_declare(self):
+        """
+        Test case to check the working of 'declare_headers'
+        function when multiple headers need to be included.
+
+        """
+        Utils.declare_headers(["tests/unit/backend/test_headers/header1.hxx",
+                               "tests/unit/backend/test_headers/header2.hxx"
+                              ])
+
+        self.assertEqual(ROOT.f(1), True)
+        self.assertEqual(ROOT.f1(2), 2)
+        self.assertEqual(ROOT.f2("myString"), "myString")
--- a/tests/unit/backend/test_headers/header2.hxx
+++ b/tests/unit/backend/test_headers/header2.hxx
@@ -1,0 +1,14 @@
+#include <string>
+
+#ifndef HEADER_2
+#define HEADER_2
+
+int f1(int x) {
+	return x;
+}
+
+std::string f2(std::string s) {
+	return s;
+}
+
+#endif
--- a/PyRDF/backend/Local.py
+++ b/PyRDF/backend/Local.py
@@ -1,5 +1,6 @@
 import ROOT
 from .Backend import Backend
+from .Utils import Utils
 
 class Local(Backend):
     """
@@ -46,8 +47,11 @@ class Local(Backend):
         Executes locally the current RDataFrame graph.
 
         """
+        from .. import includes
 
         mapper = generator.get_callable() # Get the callable
+
+        Utils.declare_headers(includes) # Declare headers if any
 
         rdf = ROOT.ROOT.RDataFrame(*generator.head_node.args) # Create RDF object
 
--- a/tests/integration/local/test_headers/header1.hxx
+++ b/tests/integration/local/test_headers/header1.hxx
@@ -1,0 +1,11 @@
+#ifndef HEADER_LOCAL_INTEGRATION_1
+#define HEADER_LOCAL_INTEGRATION_1
+
+bool check_number_less_than_5(int num){
+	/*
+	Checks if the input integer is less than 5.
+	*/
+	return num < 5;
+}
+
+#endif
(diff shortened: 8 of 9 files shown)


shravan97:
@etejedor and mentors, I've made `declare_headers` method static here as that cannot be serialized if it's an instance method

PyRDF/backend/Backend.py
```diff
@@ -52,6 +52,27 @@ def __init__(self, config={}):
 
         """
         self.config = config
+        self.includes = []
+
+    @staticmethod
+    def declare_headers(includes=[]):
```
etejedor: Why the default to []?

PyRDF/backend/Backend.py
```diff
@@ -52,6 +52,27 @@ def __init__(self, config={}):
...
+
+        parameters
+        ----------
+        includes : list (optional)
+            This list should consist of all necessary C++
+            headers as a string. This parameter defaults
+            to an empty list.
+
+        """
+        import ROOT
```
etejedor: Should we place this import at the beginning of the module?
This helper method will be executed by the backend, which anyway will import ROOT at some point because it needs to run the computation, right? So it does not hurt to have it in the module instead of in the function.

shravan97: @etejedor . Yes, that works. Will change this 👍

PyRDF/backend/Dist.py
```diff
@@ -98,6 +98,9 @@ def execute(self, generator):
         # Arguments needed to create PyROOT RDF object
         rdf_args = generator.head_node.args
 
+        # This is to avoid serialization of 'self' object
+        includes = self.includes
```
etejedor: The includes should be at main module level, not at backend level

PyRDF/__init__.py
```diff
@@ -19,4 +19,25 @@ def use(backend_name, conf = {}):
     elif backend_name == "spark":
         current_backend = Spark(conf)
     else:
-        raise Exception(" Incorrect backend environment \"{}\"".format(backend))
\ No newline at end of file
+        raise Exception(" Incorrect backend environment \"{}\"".format(backend))
+
+def include(includes_list=[]):
```
etejedor: Proposal:
- `includes_list` is renamed to `includes` and does not have a default value  
- We have a module-level variable here called `includes` (defined right after `current_backend`) which is initialized to `[]`.
- We need to support the following sequences of operations:
  - include + use
  - use + include + use
  - include + include
- To support those sequences of operations, we need that `include` appends to the module-level list the new included headers, and that the backends read that module-level list.

tests/unit/backend/test_common.py
```diff
@@ -45,3 +45,97 @@ class TestBackend(PyRDF.backend.Backend.Backend):
...
+
+    def test_default_empty_list_include(self):
+        """
+        Test case to ensure that 'PyRDF.Include' function
+        defaults to an empty list if no parameter is given.
+
+        """
+        PyRDF.current_backend = IncludeHeadersTest.TestBackend()
+
+        PyRDF.include()
```
etejedor: I would not allow this case, no default value for the list

PyRDF/__init__.py
shravan97: Yup @etejedor . I agree. That makes sense. I'll make these changes  👍

tests/unit/backend/test_common.py
shravan97: I'll remove it @etejedor 👍

PyRDF/__init__.py
shravan97: Can I retain the name of `includes_list` @etejedor ? There'd be a clash in the names otherwise

tests/unit/backend/test_common.py
```diff
@@ -45,3 +45,97 @@ class TestBackend(PyRDF.backend.Backend.Backend):
...
+    method 'declare_headers' in Backend class.
+
+    """
+    def test_single_header_declare(self):
+        """
+        Test case to check the working of 'declare_headers'
+        function when a single header needs to be included.
+
+        """
+        PyRDF.backend.Backend.Backend.declare_headers(["tests/unit/backend/test_headers/header1.cpp"])
```
etejedor: Can this be `PyRDF.current_backend.declare_headers`?

etejedor: We will also need to test this functionality with all the backends that we implement (Local, Spark), to make sure that the declared headers are taken when running a computation with that backend.

shravan97: `current_backend` refers to an instance of `Backend` @etejedor . But `declare_headers` is a static method. So that wouldn't work right ?

tests/unit/backend/test_headers/header1.cpp
```diff
@@ -0,0 +1,8 @@
+#ifndef HEADER_1
```
etejedor: This and the next one should be `.hxx`

shravan97: 👍

tests/unit/backend/test_headers/header2.cpp
```diff
@@ -0,0 +1,12 @@
+#ifndef HEADER_2
+#define HEADER_2
+
+int f1(int x) {
+	return x;
+}
+
+string f2(string s) {
```
etejedor: std::string?

shravan97: Yes @etejedor . ROOT takes the `std` namespace by default right ?

tests/unit/backend/test_common.py
shravan97: Nop sorry. My bad. It'd still work @etejedor

PyRDF/__init__.py
etejedor: Ok!

tests/unit/backend/test_common.py
shravan97: @etejedor , can I write a couple more integration tests for checking each backend after this PR ?

tests/unit/backend/test_headers/header2.cpp
etejedor: Yes that's correct, but it does not harm that your headers are complete (include + namespace) :D

shravan97: Sure @etejedor . Will change this 👍

tests/unit/backend/test_common.py
etejedor: What if then you find a problem with your code when writing the integration tests? :) I think you can include it here, just take the time that you need.

shravan97: Ok works @etejedor . In that case, can I reuse an existing integration test or tutorial and only change one function call to a header include ?

PyRDF/__init__.py
shravan97: One question @etejedor . If I declare an `includes` variable in PyRDF module level, then, how would I get the value of that variable inside a Backend class ? Is `from .. import includes` fine ?

PyRDF/backend/Backend.py
shravan97: Will remove this 👍

shravan97: One more thing here @etejedor is, if `includes` is gonna be in module-level, then I don't have to pass in a parameter here and I can just import it here directly right ?

tests/unit/backend/test_common.py
etejedor: Yes, that would work. Make a proposal and we will see :)

shravan97: Sure 👍

PyRDF/__init__.py
etejedor: Yes, or `from PyRDF import includes` should work too. Doesn't `Proxy` import `current_backend` already? Same thing here.

PyRDF/backend/Backend.py
etejedor: Mmm well, if this is the only place in `Backend` where you use `includes`, perhaps passing it as a parameter of `declare_headers` is fine.

shravan97: @etejedor , I can as well do an import inside the function `declare_headers` right ? `from .. import includes` ? Either way I have to do this import right ?

PyRDF/__init__.py
shravan97: Yup. Inside `Proxy` I'm doing a relative import. So, I have to do something similar inside `Backend` as well

PyRDF/backend/Backend.py
etejedor: Or like I said, just pass it as parameter of `declare_headers` and you do not need the import.

shravan97: @etejedor , you'll need the import to get the value of `includes` right ? Inside one of the backends, before execution, in order to get the value of `includes`, you'll have to import it right ?

etejedor: I do not get it now, but make the changes you find aproppriate, and I will re-review

shravan97: Ok @etejedor . Will do 👍

shravan97:
@etejedor , I have a proposal about making the `declare_headers` function non-static. If I declare a `__getstate__` (which simply returns an empty dict) and a `__setstate__` (which simply has a `pass`) inside `Backend` and make `Backend` serializable, I can use `declare_headers` as an instance method even while inside executors. What do you say ?

eguiraud:
what's the problem with having a class method?

shravan97:
@bluehood , not any big issue. But, I just felt it cleaner to call `self.declare_headers(...)` instead of `MyBackend.declare_headers(...)`

etejedor:
@shravan97 I have a couple of questions:
- Why does the `Backend` class need to be serializable?
- How does `declare_headers` affect the serializability of the `Backend` class?

shravan97:
@etejedor , my proposal was to call `declare_headers` as `self.declare_headers` inside the mapper. For that `self` has to be serializable. Which means it'd be better to make the parent `Backend` class itself serializable

shravan97:
Btw, @etejedor , I have to thank you for asking me to write a Spark integration test here. Because of this, I have now identified a bug in the reducer here [https://github.com/shravan97/PyRDF/blob/master/PyRDF/backend/Dist.py#L173](https://github.com/shravan97/PyRDF/blob/master/PyRDF/backend/Dist.py#L173). It seems that Python 2 has another type called `long` apart from `int`. But Python 3 doesn't have this. Which means, I must check the version inside the reducer and check if a `long` type as well exists

etejedor:
@shravan97 sure, no problem. Another option would be to have `declare_headers` defined as a class method in a `Utils` class, since its functionality is not strictly related to the backend. It is actually a utility function that allows you to inject code into the interpreter.

shravan97:
@etejedor , yes I agree. That's a great idea. Do you suggest that the `Utils` be inside `backend/` or  outside  ?

etejedor:
Let's place it inside `backend` for the moment, we can always take it out if necessary.

shravan97:
Ok @etejedor . Will do that. Btw, I have to push the Spark integration test in a separate PR and here are the 2 reasons : 
* Reducer has a bug as I pointed out earlier. Python 2 has `int` and `long` types whereas Python 3 only has `int`. And Spark returns a `long` type value which gives me a wrong output in Python 2. I've fixed this now using `numbers` module in Python though.
* There is not way to access the integer output while using Spark backend as `GetValue` would currently throw an error. There's a hacky way though : `myProxy.action_node.value`. But that doesn't sound very clean to me

shravan97:
@etejedor , made the changes now. Everything is in except an integration test for Spark. That integration test is also ready and passes. But, since it requires a change in the mapper function inside `Dist`, I'll make that a part of a separate PR. And also, the only way to access the final output value after spark execution is : `proxy.action_node.value`. Once we come up with a solution for getting the output value, I'll make the PR for that integration test

shravan97:
And @etejedor , I propose these commits after rebase : 
* One for `declare_headers` function
* One for `PyRDF.include` function
* One for changes in `Local` and `Dist` classes (invoking `declare_headers`) that is.
* One for unit tests
* One for local integration test

PyRDF/__init__.py
```diff
@@ -19,4 +20,25 @@ def use(backend_name, conf = {}):
...
+def include(includes_list):
+    """
+    Includes a list of C++ headers to be
+    declared before execution.
+
+    parameters
+    ----------
+    includes_list : list or str
+        This list should consist of all necessary C++
+        headers as a string. This parameter defaults
```
etejedor: The parameter no longer defaults as an empty list

shravan97: Will fix this @etejedor 👍

PyRDF/backend/Utils.py
```diff
@@ -0,0 +1,24 @@
...
+
+class Utils(object):
+	"""
+	Class that houses general utility
+	functions.
+
+	"""
+	@classmethod
+	def declare_headers(cls, includes):
+		"""
```
etejedor: Indentation needs fix here?

PyRDF/backend/Utils.py
```diff
@@ -0,0 +1,24 @@
...
+        PyROOT's "ROOT.gInterpreter.Declare".
+
+        parameters
+        ----------
+        includes : list
+            This list should consist of all necessary C++
+            headers as a string. This parameter defaults
+            to an empty list.
+
+		"""
```
etejedor: And here

tests/unit/backend/test_headers/header2.hxx
```diff
@@ -0,0 +1,12 @@
+#ifndef HEADER_2
+#define HEADER_2
+
```
etejedor, line 5: `#include <string>`

shravan97: ROOT interpreter already includes `string` library right @etejedor ?

shravan97: That doesn't throw an error too

tests/integration/local/test_headers_include.py
```diff
@@ -0,0 +1,28 @@
...
+
+        """
+        PyRDF.include("tests/integration/local/test_headers/header1.hxx")
+
+        rdf = PyRDF.RDataFrame(10)
+        rdf_defined = rdf.Define("b1", "(double) tdfentry_")\
+                         .Define("b2", "(int) tdfentry_ * tdfentry_")
+
+        # This filters out all numbers less than 40 in column 'b2'
+        rdf_filtered = rdf_defined.Filter("check_number_less_than_40(b2)")
```
etejedor: To make this test simpler, what about just `rdf_defined.Filter("check_number_less_than_5(tdfentry_)")`, and you do not need the previous Defines.

tests/unit/backend/test_common.py
```diff
@@ -45,3 +46,87 @@ class TestBackend(PyRDF.backend.Backend.Backend):
 
         with self.assertRaises(TypeError):
             TestBackend()
+
+class IncludeHeadersTest(unittest.TestCase):
+    """
+    Tests to check the working of 'PyRDF.Include' function.
```
etejedor: It is `PyRDF.include`, lowercase

tests/integration/local/test_headers_include.py
shravan97: Yup @etejedor . That is better. Will make this fix 👍

tests/unit/backend/test_common.py
shravan97: 👍

tests/unit/backend/test_common.py
```diff
@@ -45,3 +46,87 @@ class TestBackend(PyRDF.backend.Backend.Backend):
...
+        self.assertEqual(ROOT.f(1), True)
+
+    def test_multiple_headers_declare(self):
+        """
+        Test case to check the working of 'declare_headers'
+        function when multiple headers need to be included.
+
+        """
+        Utils.declare_headers(["tests/unit/backend/test_headers/header1.hxx",
+                                                       "tests/unit/backend/test_headers/header2.hxx"
```
etejedor: Check indentation here

PyRDF/backend/Utils.py
```diff
@@ -0,0 +1,24 @@
...
+	def declare_headers(cls, includes):
+		"""
+        Declares all required headers using
+        PyROOT's "ROOT.gInterpreter.Declare".
+
+        parameters
+        ----------
+        includes : list
+            This list should consist of all necessary C++
+            headers as a string. This parameter defaults
```
etejedor: No default anymore

tests/unit/backend/test_headers/header2.hxx
etejedor: Yes, it includes it and it also makes the std namespace available. But I prefer to write complete headers that also compile without the ROOT interpreter :)

shravan97: Ok @etejedor . Will make this change 👍

etejedor:
Finished my review, after the last requested changes, this can go in. Ok for the structure of commits proposed.

shravan97:
@etejedor , changes are done. Can I rebase now ?

etejedor:
Yes please

shravan97:
@etejedor , rebase done and ready to be merged 👍

PR #49, opened 2018-07-29
Implement duck typing in Proxy and modify tutorials and tests

shravan97:
In this PR : 
* Change Local execution to store actual values instead of RResultPtrs
* Implement duck typing Proxy instances
* Add tests for the new Proxy features
* Modify tests and tutorials to remove `GetValue` calls

### DIFF
--- a/PyRDF/Proxy.py
+++ b/PyRDF/Proxy.py
@@ -34,6 +34,184 @@ class Proxy(object):
         """
         self.action_node = action_node
 
+    def __eq__(self, obj):
+        """
+        Checks if the given object is equal to
+        the value of the current action node.
+
+        Parameters
+        ----------
+        obj : numbers.Number or PyRDF.Proxy.Proxy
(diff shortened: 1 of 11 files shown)


shravan97:
@etejedor , @diocas and @bluehood , I've not added a `GetValue` method in `Proxy` as we discussed. But, I'm concerned about cases where a user might require a reference to final value. If the user's needs to carry out an operation that cannot be done on the current implementation of `Proxy`, then, the best way would be to use a `GetValue`. The current implementation of Proxy only supports : 
* Addition
* subtraction
* Division
* Multiplication
* Equal to (`==`)
* `str()`

etejedor:
@shravan97 let's discuss this PR on Friday at our meeting. I am not convinced we should merge these changes, but let's decide it all together.

shravan97:
Sure @etejedor :+1:

PR #50, opened 2018-08-03
Change Local to store actual values and implement GetValue in Proxy 

shravan97:
In this PR : 
* Change Local implementation to store the actual values instead of RResultPtrs
* Implement `GetValue` method in `Proxy`
* Write tests for `GetValue` method

### DIFF
--- a/PyRDF/Proxy.py
+++ b/PyRDF/Proxy.py
@@ -3,11 +3,10 @@ from .CallableGenerator import CallableGenerator
 
 class Proxy(object):
     """
-    This class wraps action nodes and instances of Proxy act as
-    futures of the result produced by some action. They implement
-    a lazy synchronization mechanism, i.e., when they are accessed
-    for the first time, they trigger the execution of the whole
-    RDataFrame graph.
+    Instances of Proxy act as futures of the result produced
+    by some action. They implement a lazy synchronization
+    mechanism, i.e., when they are accessed for the first time,
+    they trigger the execution of the whole RDataFrame graph.
 
     Attributes
     ----------
@@ -50,13 +49,23 @@ class Proxy(object):
         self._cur_attr = attr # Stores the name of operation call
         return self._call_handler
     
-    def _call_handler(self, *args, **kwargs):
-        # Handles an operation call to the current action node
-        # and returns result of the current action node.
+    def GetValue(self):
+        """
+        Returns the result value of the current action
+        node if it was executed before, else triggers
+        the execution of the entire PyRDF graph before
+        returning the value.
 
+        """
         if not self.action_node.value: # If event-loop not triggered
             from . import current_backend
             generator = CallableGenerator(self.action_node.get_head())
             current_backend.execute(generator)
 
-        return getattr(self.action_node.value, self._cur_attr)(*args, **kwargs)
+        return self.action_node.value
+
+    def _call_handler(self, *args, **kwargs):
+        # Handles an operation call to the current action node
+        # and returns result of the current action node.
+
+        return getattr(self.GetValue(), self._cur_attr)(*args, **kwargs)
--- a/PyRDF/backend/Local.py
+++ b/PyRDF/backend/Local.py
@@ -63,6 +63,10 @@ class Local(Backend):
         values[0].GetValue() # Trigger event-loop
 
         for i in range(len(values)):
-            # Set the values of action nodes to
-            # 'TResultPtr's obtained
-            nodes[i].value = values[i]
+            # Set the obtained values and
+            # 'RResultPtr's of action nodes
+            nodes[i].value = values[i].GetValue()
+            # We store the 'RResultPtr's because,
+            # those should be in scope while doing
+            # a 'GetValue' call on them
+            nodes[i].ResultPtr = values[i]
--- a/tests/unit/test_proxy.py
+++ b/tests/unit/test_proxy.py
@@ -1,6 +1,8 @@
 from PyRDF.Proxy import Proxy
 from PyRDF.Node import Node
-import unittest
+from PyRDF.backend.Backend import Backend
+from PyRDF import RDataFrame
+import unittest, PyRDF
 
 class AttrReadTest(unittest.TestCase):
     """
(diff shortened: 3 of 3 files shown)


eguiraud:
@shravan97 action nodes and RResultPtrs are two different things, please clear the confusion in the variable naming and documentation.

```python
r = df.Action()
```
creates an action node, adds it to the graph, then creates a RResultPtr connected to the action and returns the RResultPtr.

`GetValue` is a method of `RResultPtr`, not of an action node.

eguiraud:
If I call
```python
dff = df.Filter("true")
```
in PyRDF that returns a `Proxy` to a node, is that correct? If yes, there is a problem: `dff` cannot have a `GetValue` method.

shravan97:
No @bluehood . `df.Filter("true")` returns an object of `Node` type. If the user calls `GetValue` on this type of object, there would be an error during the execution of event-loop anyway

shravan97:
And @bluehood , `df.Action()` isn't implemented yet in PyRDF. For this, I have to add `Action` as a supported operation

eguiraud:
`df.Action` was just an example with a placeholder name.

Ok my mistake on `Proxy`, although I hope the user gets an error immediately if they call `GetValue` on a node, not during the event loop.

Thank you for the clarifications, so my comment becomes: please clarify in the documentation of `Proxy` that what it proxies is a result, not an action, and `GetValue` will be called on the `RResultPtr`, not on the action node.

shravan97:
@bluehood , the current implementation throws an error immediately if user calls `GetValue` on a `Node` object. But error message is quite generic though. It'd just raise that `The current backend doesn't support "GetValue"`. Besides, I'm not explicitly checking if `GetValue` is called on a `Node` object. It throws an error because, `GetValue` is not in the supported operations. Is that fine ?

shravan97:
Also @bluehood , technically here, `GetValue` is not called on a RResultPtr. I've changed the implementation now to store the actual values itself instead of RResultPtrs as we discussed in the meeting

eguiraud:
> Is that fine?

yes, that's consistent.

>  technically here, GetValue is not called on a RResultPtr

you're right, I take it back

shravan97:
But yes @bluehood , I'll add a comment to make it clear for the users :+1:

PyRDF/backend/Local.py
```diff
@@ -63,6 +63,9 @@ def execute(self, generator):
...
-            # Set the values of action nodes to
-            # 'TResultPtr's obtained
-            nodes[i].value = values[i]
\ No newline at end of file
+            # The type of the value of the action node
+            value_type = type(values[i].GetValue())
+            # The `value_type` is required here because,
+            # after a call to `GetValue`, the values die
+            # along with the RResultPtrs
+            nodes[i].value = value_type(values[i].GetValue())
```
etejedor: Why do you need to copy construct it? Can't you just do:
`nodes[i].value = values[i].GetValue()` ?

PyRDF/Proxy.py
```diff
@@ -50,13 +50,23 @@ def __getattr__(self, attr):
         self._cur_attr = attr # Stores the name of operation call
         return self._call_handler
     
-    def _call_handler(self, *args, **kwargs):
-        # Handles an operation call to the current action node
-        # and returns result of the current action node.
+    def GetValue(self):
+        """
+        Returns the value of the current action node
```
etejedor: Returns the result value of the current action node ...

PyRDF/backend/Local.py
shravan97: @etejedor , the same reason as spark backend. It seems to throw the same error workout that

etejedor: Is it because the `RResultPtr`s are going out of scope and thus garbarge collected, together with their values?
In Spark we need to copy construct because we cannot serialize `RResultPtrs`. In the local backend, we do not need to serialize them, so if we want to avoid the copy we just need to keep a reference to them somewhere.

shravan97: @etejedor yes. `RResultPtr`s are going out of scope without the copy construction I believe. This is the error that pops up without copy construction : `AttributeError: 'PyROOT_NoneType' object has no attribute 'Draw' `. That is for calling a `Draw` on a Histo object

shravan97: @etejedor , I'm guessing that this is because I'm constructing the rdf object and triggering the event-loop within the `execute` method of Local. So, may be `RResultPtr`s get out of scope after execution when we return back to the `Proxy` object from where `execute` was called

shravan97: @etejedor and @bluehood , any fix for this ?

etejedor: To avoid the copy in the local backend case, I suggest we keep storing the `RResultPtr`s. Then, in the logic of `GetValue` in `Proxy`, we can invoke `GetValue` on the stored object if it is an `RResultPtr`, otherwise we just returned the stored value.
This can also be implemented with a `try` `except`, returning `GetValue` in the try and returning the stored value in the `except`.

shravan97: @etejedor , but then it wouldn't be consistent across all backends then. Is that okay ?

etejedor: Another option is to store both the RResultPtr and the actual values. In the case of the Spark backend, it would only set the actual values. In the case of the local backend, it would set both. Keeping the RResultPtrs would just prevent the values from disappearing.

shravan97: @etejedor , I don't understand this. Could you elaborate ?

shravan97: @etejedor , resolved this now 👍 
Sorry for the delay

shravan97:
@etejedor , I propose 3 commits for this PR after rebase : 
* One for changes in `Local.py`
* One for `GetValue` method in `Proxy.py`
* One for `GetValue` unit tests
Let me know once you're done with the review, so that I can rebase this

etejedor:
@shravan97 please rearrange the commits.

shravan97:
@etejedor , done rebasing now 👍

PR #51, opened 2018-08-03
Raise an exception if ROOT.gInterpreter.Declare returned false

shravan97:
In this PR : 
* Throw an error if `ROOT.gInterpreter.Declare(...)` returned false

### DIFF
--- a/PyRDF/backend/Utils.py
+++ b/PyRDF/backend/Utils.py
@@ -20,4 +20,5 @@ class Utils(object):
 
         """
         for header in includes:
-            ROOT.gInterpreter.Declare("#include \"{}\"\n".format(header))
+            if not ROOT.gInterpreter.Declare("#include \"{}\"\n".format(header)):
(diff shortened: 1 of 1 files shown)


PyRDF/backend/Utils.py
```diff
@@ -20,4 +20,5 @@ def declare_headers(cls, includes):
 
         """
         for header in includes:
-            ROOT.gInterpreter.Declare("#include \"{}\"\n".format(header))
\ No newline at end of file
+            if not ROOT.gInterpreter.Declare("#include \"{}\"\n".format(header)):
+                raise Exception("There was an error in including \"{}\". Please fix the error and retry !".format(header))
```
etejedor: I would suppress the "Please fix the error and retry" part of the message. More information will be printed by cling.

shravan97:
This is done @etejedor . Rebased as well, so that we don't spend a lot of time on this

PR #52, opened 2018-08-04
Remove the logic for processing integers in reducer and add Spark tutorials

shravan97:
In this PR : 
* Remove support for operations on integers like "Sum", "Mean", "Max"....etc. to wait for the generic reducer
* Add tutorials for Spark
* Modify `.travis-script.sh` and `run_tests.sh` to run Spark tutorials as well

### DIFF
--- a/PyRDF/backend/Dist.py
+++ b/PyRDF/backend/Dist.py
@@ -24,6 +24,11 @@ class Dist(Backend):
         super(Dist, self).__init__(config)
         # Operations that aren't supported in distributed backends
         operations_not_supported = [
+        'Sum',
+        'Mean',
+        'Max',
+        'Min',
+        'Count',
         'Range',
         'Take',
         'Snapshot',
@@ -175,9 +180,19 @@ class Dist(Backend):
                 if isinstance(values_list1[i], ROOT.TH1) or isinstance(values_list1[i], ROOT.TH2):
                     # Merging two objects of type ROOT.TH1D or ROOT.TH2D
                     values_list1[i].Add(values_list2[i])
-                elif isinstance(values_list1[i], int):
-                    # Merging two objects of type `int`
-                    values_list1[i] += values_list2[i]
+                elif isinstance(values_list1[i], ROOT.TGraph):
+                    # Prepare a TList
+                    tlist = ROOT.TList()
+                    tlist.Add(values_list2[i])
+
+                    # Merge the second graph onto the first
+                    num_points = values_list1[i].Merge(tlist)
+
+                    # Check if there was an error in merging
+                    if num_points == -1:
+                        raise Exception("Error reducing two result values of type TGraph !")
+                else:
+                    raise NotImplementedError("The type \"{}\" isn't supported by the reducer yet !".format(type(values_list1[i])))
 
             return values_list1
 
--- a/tests/integration/spark/test_reducer_merge.py
+++ b/tests/integration/spark/test_reducer_merge.py
@@ -1,0 +1,168 @@
+import unittest, PyRDF, ROOT
+
+class ReducerMergeTest(unittest.TestCase):
+    """
+    Integration tests to check the working of merge operations
+    in the reducer function.
+
+    """
+    @classmethod
+    def setUpClass(cls):
+        """
+        Set up method to select Spark backend
+        before running all the tests.
+
+        """
+        PyRDF.use("spark", {'npartitions':2, 'spark.executor.instances':2})
+
+    def assertHistoOrProfile(self, obj_1, obj_2):
+        """
+        Asserts equality between two 'ROOT.TH1' or
+        'ROOT.TH2' objects.
+
+        """
+        # Compare the sizes of equivalent objects
+        self.assertEqual(obj_1.GetEntries(), obj_2.GetEntries())
+
+        # Compare the means of equivalent objects
+        self.assertEqual(obj_1.GetMean(), obj_2.GetMean())
+
+        # Compare the standard deviations of equivalent objects
+        self.assertEqual(obj_1.GetStdDev(), obj_2.GetStdDev())
+
+    def define_two_columns(self, rdf):
+        """
+        Helper method that Defines and returns
+        two columns with definitions "x = tdfentry_"
+        and "y = tdfentry_ * tdfentry_".
+
+        """
+        return rdf.Define("x", "tdfentry_").Define("y", "tdfentry_*tdfentry_")
+
+    def define_three_columns(self, rdf):
+        """
+        Helper method that Defines and returns three columns
+        with definitions "x = tdfentry_", "y = tdfentry_ * tdfentry_"
+        and "z = tdfentry_ * tdfentry_ * tdfentry_".
+
+        """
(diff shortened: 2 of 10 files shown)


PyRDF/backend/Dist.py
```diff
@@ -175,9 +175,8 @@ def reducer(values_list1, values_list2):
                 if isinstance(values_list1[i], ROOT.TH1) or isinstance(values_list1[i], ROOT.TH2):
                     # Merging two objects of type ROOT.TH1D or ROOT.TH2D
                     values_list1[i].Add(values_list2[i])
-                elif isinstance(values_list1[i], int):
-                    # Merging two objects of type `int`
-                    values_list1[i] += values_list2[i]
+                else:
+                    raise NotImplementedError("The reducer function doesn't support operations that give a integer output yet !")
```
etejedor: The `else` here doesn't just cover integers, I would just say that `type(values_list1[i]` cannot be merged.
On the other hand, we need to be able to merge  more things here, as discussed in the previous meeting. `TProfile`s are covered by `TH1` (it is a subclass), but we still need `TGraph`. Can you add a check here for the `TGraph` type? In order to merge, you can use:
https://root.cern.ch/doc/v614/classTGraph.html#a10c31c206c4559803156b9b4a4b8b997

shravan97:
@etejedor , made the changes, wrote tests and rebased as well 👍

PyRDF/backend/Dist.py
```diff
@@ -175,9 +180,19 @@ def reducer(values_list1, values_list2):
...
+                    tlist.Add(values_list2[i])
+
+                    # Merge the second graph onto the first
+                    num_points = values_list1[i].Merge(tlist)
+
+                    # Check if there was an error in merging
+                    if num_points == -1:
+                        raise Exception("There was an error in merging these 2 TGraphs !")
+                else:
+                    raise NotImplementedError("The type \"{}\" isn't supported by the reducer yet !")
```
etejedor: You need to append `.format` here for the string replacement of the type

shravan97:
Fixed and rebased @etejedor 👍

PyRDF/backend/Dist.py
```diff
@@ -175,9 +180,19 @@ def reducer(values_list1, values_list2):
...
+                    # Prepare a TList
+                    tlist = ROOT.TList()
+                    tlist.Add(values_list2[i])
+
+                    # Merge the second graph onto the first
+                    num_points = values_list1[i].Merge(tlist)
+
+                    # Check if there was an error in merging
+                    if num_points == -1:
+                        raise Exception("There was an error in merging these 2 TGraphs !")
```
etejedor: I would say: "Error reducing two result values of type TGraph"

tutorials/spark/df003_profiles.py
```diff
@@ -0,0 +1,50 @@
...
+## RDataFrame. See the documentation of TProfile and TProfile2D to better
+## understand the analogy of this code with the example one.
+##
+## \macro_code
+##
+## \date February 2017
+## \author Danilo Piparo
+
+import ROOT, PyRDF
+PyRDF.use("spark")
```
etejedor: Should we show in some tutorial how to configure some Spark parameters, or the number of partitions?

tests/integration/spark/test_reducer_merge.py
```diff
@@ -0,0 +1,164 @@
+import unittest, PyRDF, ROOT
+
+class ReducerMergeTest(unittest.TestCase):
+    """
+    Integration tests to check the working of merge operations
+    in the reducer function.
+
+    """
+    def setUp(self):
```
etejedor: Since you need this for all the tests and the backend is a singleton, this should be
```python
@classmethod
def setUpClass(cls):
```
so that it is executed only once for all the tests in this class.

tests/integration/spark/test_reducer_merge.py
```diff
@@ -0,0 +1,164 @@
...
+        """
+        Set up method to select Spark backend
+        for every test.
+
+        """
+        PyRDF.use("spark")
```
etejedor: What are the defaults for number of partitions and number of workers in Spark? Are we sure things are done in parallel? Perhaps setting it explicitly would be better for the clarity of the test.

tests/integration/spark/test_reducer_merge.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def test_histo1d_merge(self):
+        """
+        Integration test to check the working of
+        Histo1D merge operation in the reducer.
+
+        """
```
etejedor, line 56: There is quite some repetition in these tests. Can we refactor the common part into helper functions and just leave the code that is specific to every test (basically, the creation of the histos/profiles).

tests/integration/spark/test_reducer_merge.py
```diff
@@ -0,0 +1,164 @@
...
+        # Operations with PyROOT
+        rdf_cpp = ROOT.ROOT.RDataFrame(10)
+        columns_cpp = rdf_cpp.Define("x", "tdfentry_").Define("y", "tdfentry_*tdfentry_")
+        graph_cpp = columns_cpp.Graph("x", "y")
+
+        # Sort the graphs to make sure corresponding points are same
+        graph_py.Sort()
+        graph_cpp.Sort()
+
+        # Compare the means of equivalent graphs
```
etejedor: These are not the means

tests/integration/spark/test_reducer_merge.py
```diff
@@ -0,0 +1,164 @@
...
+        self.assertListEqual(list(graph_py.GetX()), list(graph_cpp.GetX()))
+
+        # Compare the covariances of equivalent graphs
```
etejedor: These are not the covariances

tutorials/spark/df003_profiles.py
shravan97: Yup @etejedor . It'd be better if I include that in one of the tutorials. I'll do it in this tutorial (`df003_profiles.py`) 👍

shravan97:
@etejedor , made the changes and rebased 👍

tests/integration/spark/test_reducer_merge.py
```diff
@@ -0,0 +1,168 @@
...
+        return rdf.Define("x", "tdfentry_").Define("y", "tdfentry_*tdfentry_").Define("z", "tdfentry_*tdfentry_*tdfentry_")
+
+    def test_histo1d_merge(self):
+        """
+        Integration test to check the working of
+        Histo1D merge operation in the reducer.
+
+        """
+        # Operations with PyRDF
+        rdf_py = PyRDF.RDataFrame(10)
```
etejedor, line 58: What about refactoring a bit more:
```python
rdf_py, rdf_cpp = self.create_test_rdfs()
```
Same below.

shravan97: But @etejedor , wouldn't it be more clear if I separate the declaration of `PyRDF` and `PyROOT` variables ?

shravan97: And it also separately defines what I want to test and what I want to assert it with. What do you say ?

etejedor: Ok let's merge

PR #55, opened 2018-08-10
Prune action nodes that were already computed and add demo

shravan97:
In this PR : 
* Prune action nodes that were already computed
* Add unit tests for the pruning
* Add a demo notebook


PyRDF/Node.py
```diff
@@ -194,4 +194,13 @@ def graph_prune(self):
             
             return False
 
+        elif not self.children and self.value:
```
etejedor: `not self.children` is checked twice in this if-elif, can you reorganize the conditions to check self.children only once at the beginning?

PyRDF/Node.py
```diff
@@ -194,4 +194,13 @@ def graph_prune(self):
...
+
+            ### If the current node's value was already
+            ### computed, it should get pruned only if it's
+            ### an action node.
+
+            if self.operation and self.operation.is_action():
```
etejedor: Isn't this condition doing the contrary than what one would expect? Here you are **not** pruning if it is an action that already has a value, and that is precisely what you **do** want to prune, right?

shravan97: @etejedor , there's actually some confusion here. It looks like I'm pruning the nodes only when `graph_prune` returns `false`. But I've documented it the opposite way. I'm fixing this right away 👍

shravan97:
@etejedor , changes made and rebased too 👍

